PR #2, opened 2021-02-23
bx added

pyTony:
I hope it is good enough, I am novice with Javascript/Typescript


urish:
Thank you! We'll look at it during the live-coding stream today :)

urish:
Thanks Tony! This is now part of bcbbdf0ae1ac3effadfe6f6981abcd868a3f35de

PR #3, opened 2021-03-20
implemented sxtb, asrs, eors

BrianPopeck:
First time working with Typescript, let me know if I need to fix anything!


urish:
Thanks again for the pull request Brian!

Eventually, we did find a small bug in ASRS during the live stream, you can see the fix here: c310648b4705abd99245a618b547c783ac5283b0

Great job nevertheless

PR #23, opened 2021-05-12
test(rp2040): complete test opcode

Turro75:
just realized there was a test file to check that the generated opcodes are correct.
Added the missing ones.


urish:
Thank you valerio!

PR #26, opened 2021-05-14
Fix(rp2040): fix SUBS register N flag handling

Turro75:
gdbdiff looks like a powerful tools!


urish:
😍

urish:
Released this fix as v 0.3.8

Issue #24, opened 2021-05-13
printf() incorrectly formats numbers (Arduino Mbed Core)


urish:
To reproduce, compile the following sketch using the Arduino Mbed Core for RP2040:

```cpp
void setup() {
  printf("12 is %d\n", 12);
}

void loop() {}
```

Then run the result in the emulator. This will print to the serial console: 

    12 is <

Instead of the expected result, `12 is 12`.

urish:
Traced the issue to our LSRS (register) imlpementation.

Pre-compiled HEX file (of the above) sketch that reproduces the issue: [test-printf.ino.hex](https://github.com/wokwi/rp2040js/files/6477503/test-printf.ino.hex.txt) 

Unit test case (passes on the actual silicone, fails in the emulator):

```typescript
  it('should return zero for `lsrs r2, r3` with 32 bit shift', async () => {
    await cpu.setPC(0x20000000);
    await cpu.writeUint16(0x20000000, opcodeLSRSreg(r2, r3));
    await cpu.setRegisters({ r2: 10, r3: 32 });
    await cpu.singleStep();
    const registers = await cpu.readRegisters();
    expect(registers.r2).toEqual(0);
    expect(registers.Z).toEqual(true);
    expect(registers.C).toEqual(false);
  });
```

PR #27, opened 2021-05-14
feat(gdbdiff): print less and more readable output

Turro75:
reworked a bit gdbdiff to produce less output:

Successfully compared 0 instructions
Successfully compared 500 instructions
Successfully compared 1000 instructions
Successfully compared 1500 instructions
Successfully compared 2000 instructions
Successfully compared 2500 instructions
Successfully compared 3000 instructions
Successfully compared 3500 instructions
Successfully compared 4000 instructions
Successfully compared 4500 instructions
Successfully compared 5000 instructions
Mismatch register 3: emulator 0x0 != silicone 0x197
status register at failed instruction:
Registers 	 Emulator 	     Silicone
r0     		0x00000000		 0x00000000
r1     		0x2000cd07		 0x2000cd07
r2     		0x40034000		 0x40034000
r3     		0x2000cd07		 0x2000cd07
r4     		0x00000000		 0x00000000
r5     		0x00000001		 0x00000001
r6     		0x2000cd2f		 0x2000cd2f
r7     		0x100137d4		 0x100137d4
r8     		0x00000000		 0x00000000
r9     		0x00000000		 0x00000000
r10     		0x00000000		 0x00000000
r11     		0x00000000		 0x00000000
r12     		0x2000267d		 0x2000267d
sp     		0x2000cd00		 0x2000cd00
lr     		0x10008fa5		 0x10008fa5
pc     		0x1000b746		 0x1000b746
xPSR     		0x61000000		 0x61000000
Status register at current instruction:
Registers 	 Emulator 	     Silicone
r0     		0x00000000		 0x00000000
r1     		0x2000cd07		 0x2000cd07
r2     		0x40034000		 0x40034000
r3     		0x00000000		 0x00000197
r4     		0x00000000		 0x00000000
r5     		0x00000001		 0x00000001
r6     		0x2000cd2f		 0x2000cd2f
r7     		0x100137d4		 0x100137d4
r8     		0x00000000		 0x00000000
r9     		0x00000000		 0x00000000
r10     		0x00000000		 0x00000000
r11     		0x00000000		 0x00000000
r12     		0x2000267d		 0x2000267d
sp     		0x2000cd00		 0x2000cd00
lr     		0x10008fa5		 0x10008fa5
pc     		0x1000b748		 0x1000b748
xPSR     		0x61000000		 0x61000000

### DIFF
--- a/debug/gdbdiff.ts
+++ b/debug/gdbdiff.ts
@@ -15,13 +15,12 @@ function printComparedRegisters(
   silicone: Uint32Array
 ) {
   for (let i = 0; i < registerNames.length; i++) {
-    let modified = '';
+    let modified = ' ';
     if (emulator[i] !== silicone[i]) {
       modified = '*';
     }
     console.log(
-      registerNames[i],
-      modified,
+      registerNames[i] + modified,
       '\t\t0x' +
         dumpUint32(registers[i]) +
         '\t0x' +
@@ -50,10 +49,15 @@ function printFlags(xpsr: number) {
   return `[${negative}${zero}${carry}${overflow}]`;
 }
 
-function compareRegisters(emulator: Uint32Array, silicone: Uint32Array) {
+async function compareFixRegisters(
+  tofixclient: GDBClient,
+  emulator: Uint32Array,
+  silicone: Uint32Array
+) {
   let result = true;
   for (let i = 0; i < emulator.length; i++) {
     if (emulator[i] !== silicone[i]) {
+      await tofixclient.writeRegister(i, silicone[i]);
       result = false;
     }
   }
@@ -69,23 +73,21 @@ async function main() {
   await client1.writeRegister(19, 1, 8);
   await client2.writeRegister(19, 1, 8);
   // Start diffing
-  let lastRegSet1 = await client1.readRegisters();
-  let lastRegSet2 = await client2.readRegisters();
+  let prevRegSet = await client2.readRegisters();
   let counter = 1;
   for (counter = 1; ; counter++) {
     const regSet1 = await client1.readRegisters();
     const regSet2 = await client2.readRegisters();
 
-    if (!compareRegisters(regSet1, regSet2)) {
-      console.log('\nMismatch after ', counter, ' compared instructions');
+    if (!(await compareFixRegisters(client1, regSet1, regSet2))) {
+      console.log('\n\nMismatch after ', counter, ' compared instructions');
       console.log('\nRegister\tStartValue\tEmulator\tSilicone');
-      printComparedRegisters(lastRegSet1, regSet1, regSet2);
-      process.exit(1);
+      printComparedRegisters(prevRegSet, regSet1, regSet2);
     }
-    lastRegSet1 = regSet1;
+    prevRegSet = regSet1;
     await client1.singleStep();
     await client2.singleStep();
-    if (counter % 500 === 0) {
+    if (counter % 50 === 0) {
       console.log(`Successfully compared ${counter} instructions`);
     }
   }
--- a/debug/gdbdiff.ts
+++ b/debug/gdbdiff.ts
@@ -9,25 +9,51 @@
 
 import { GDBClient, dumpUint32, registerNames } from '../src/utils/gdbclient';
 
-function printComparedRegisters(emulator: Uint32Array, silicone: Uint32Array) {
-  console.log('Registers \t Emulator \t     Silicone');
+function printComparedRegisters(
+  registers: Uint32Array,
+  emulator: Uint32Array,
+  silicone: Uint32Array
+) {
   for (let i = 0; i < registerNames.length; i++) {
+    let modified = '';
+    if (emulator[i] !== silicone[i]) {
+      modified = '*';
+    }
     console.log(
       registerNames[i],
-      '    \t\t0x' + dumpUint32(emulator[i]) + '\t\t 0x' + dumpUint32(silicone[i])
+      modified,
+      '\t\t0x' +
+        dumpUint32(registers[i]) +
+        '\t0x' +
+        dumpUint32(emulator[i]) +
+        '\t0x' +
+        dumpUint32(silicone[i])
     );
+    if (registerNames[i] === 'xPSR' && modified === '*') {
+      console.log(
+        'Flags\t\t',
+        printFlags(registers[i]),
+        '\t',
+        printFlags(emulator[i]),
+        '\t',
+        printFlags(silicone[i])
+      );
+    }
   }
+}
+
+function printFlags(xpsr: number) {
+  let negative = xpsr & 0x80000000 ? 'N' : '-';
+  let zero = xpsr & 0x40000000 ? 'Z' : '-';
+  let carry = xpsr & 0x20000000 ? 'C' : '-';
+  let overflow = xpsr & 0x10000000 ? 'O' : '-';
+  return `[${negative}${zero}${carry}${overflow}]`;
 }
 
 function compareRegisters(emulator: Uint32Array, silicone: Uint32Array) {
   let result = true;
   for (let i = 0; i < emulator.length; i++) {
     if (emulator[i] !== silicone[i]) {
-      console.log(
-        `Mismatch register ${i}: emulator 0x${emulator[i].toString(16)} != silicone 0x${silicone[
-          i
-        ].toString(16)}`
-      );
       result = false;
     }
   }
@@ -45,26 +71,23 @@ async function main() {
   // Start diffing
   let lastRegSet1 = await client1.readRegisters();
   let lastRegSet2 = await client2.readRegisters();
-  let counter = 0;
-  for (counter = 0; ; counter++) {
+  let counter = 1;
+  for (counter = 1; ; counter++) {
     const regSet1 = await client1.readRegisters();
     const regSet2 = await client2.readRegisters();
-    //await client1.dumpRegisters();
+
     if (!compareRegisters(regSet1, regSet2)) {
-      console.log('status register at failed instruction:');
-      printComparedRegisters(lastRegSet1, lastRegSet2);
-      console.log('Status register at current instruction:');
-      printComparedRegisters(regSet1, regSet2);
-      break;
+      console.log('\nMismatch after ', counter, ' compared instructions');
+      console.log('\nRegister\tStartValue\tEmulator\tSilicone');
+      printComparedRegisters(lastRegSet1, regSet1, regSet2);
+      process.exit(1);
     }
     lastRegSet1 = regSet1;
-    lastRegSet2 = regSet2;
     await client1.singleStep();
     await client2.singleStep();
     if (counter % 500 === 0) {
-      console.log('Successfully compared ' + counter + ' instructions');
+      console.log(`Successfully compared ${counter} instructions`);
     }
-    //console.log('---');
   }
 }
 
--- a/debug/gdbdiff.ts
+++ b/debug/gdbdiff.ts
@@ -7,13 +7,27 @@
  * Copyright (C) 2021, Uri Shaked.
  **/
 
-import { GDBClient } from '../src/utils/gdbclient';
+import { GDBClient, dumpUint32, registerNames } from '../src/utils/gdbclient';
+
+function printComparedRegisters(emulator: Uint32Array, silicone: Uint32Array) {
+  console.log('Registers \t Emulator \t     Silicone');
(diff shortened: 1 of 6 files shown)


Turro75:
Sorry Uri, I don't know why github still remember the previous commits even if You already merged them.
Can I ask You how I can avoid this annoying behavior?

The only relevant commit is b0b6093

urish:
Hi Valerio, 

First of all, this is a great idea, thanks! I have some comments, I will leave a detailed review later today.

Now for your question, you can try to rebase your branch against the current master:

1. `git add remote upstream https://github.com/wokwi/rp2040js` (only do it once, git will remember it for the feature)
2. `git fetch upstream`
3. `git rebase upstream/master`
4. if you already pushed the changes (like in this case), force push them with `git push --force-with-lease`

Also, it's a good practice to open a new branch per pull-request. Whenever you update the branch, the pull-request is also updated automatically

debug/gdbdiff.ts
```diff
@@ -7,13 +7,27 @@
...
+import { GDBClient, dumpUint32, registerNames } from '../src/utils/gdbclient';
+
+function printComparedRegisters(emulator: Uint32Array, silicone: Uint32Array) {
+  console.log('Registers \t Emulator \t     Silicone');
+  for (let i = 0; i < registerNames.length; i++) {
+    console.log(
+      registerNames[i],
+      '    \t\t0x' + dumpUint32(emulator[i]) + '\t\t 0x' + dumpUint32(silicone[i])
+    );
+  }
```
urish, line 19: Also, it will be helpful if we print the individual APSR flag values (N/Z/C/V), that's where most of the problems are.

I think we can just print a list of the flags with are on, e.g. if only N and C are true, we'll print `[N-C-]` ,and if only Z is true we'll print `[-Z--]`, and if all flags are empty `[----]`

debug/gdbdiff.ts
```diff
@@ -23,25 +37,34 @@ function compareRegisters(emulator: Uint32Array, silicone: Uint32Array) {
...
-  for (let counter = 0; ; counter++) {
+  let lastRegSet1 = await client1.readRegisters();
+  let lastRegSet2 = await client2.readRegisters();
+  let counter = 0;
+  for (counter = 0; ; counter++) {
     const regSet1 = await client1.readRegisters();
     const regSet2 = await client2.readRegisters();
-    console.log(`Instruction ${counter}`);
-    await client1.dumpRegisters();
+    //await client1.dumpRegisters();
```
urish, line 52: no need to keep this line!

debug/gdbdiff.ts
```diff
@@ -23,25 +37,34 @@ function compareRegisters(emulator: Uint32Array, silicone: Uint32Array) {
...
     }
+    lastRegSet1 = regSet1;
+    lastRegSet2 = regSet2;
     await client1.singleStep();
     await client2.singleStep();
-    console.log('---');
+    if (counter % 500 === 0) {
+      console.log('Successfully compared ' + counter + ' instructions');
+    }
+    //console.log('---');
```
urish, line 67: no need to keep this line too

urish (changes requested):
Thanks for coming up with this idea! I've left some suggestions and requests, please feel free to ask any questions if anything is not clear (or if you have a different approach).

Turro75:
now the output is this

 [diff.txt](https://github.com/wokwi/rp2040js/files/6481563/diff.txt)

flags are shown only if xPSR has mismatch

debug/gdbdiff.ts
```diff
@@ -23,25 +63,31 @@ function compareRegisters(emulator: Uint32Array, silicone: Uint32Array) {
...
-  await client2.connect('raspberrypi');
+  await client1.connect('localhost', 3334);
+  await client2.connect('localhost', 3333);
   // Disable interrupts
   await client1.writeRegister(19, 1, 8);
   await client2.writeRegister(19, 1, 8);
   // Start diffing
-  for (let counter = 0; ; counter++) {
+  let lastRegSet1 = await client1.readRegisters();
+  let lastRegSet2 = await client2.readRegisters();
```
urish, line 73: This is unused, so we we remove it (and possibly also rename `lastRegSet1` to `lastRegSet` or maybe `prevRegSet`)

urish (changes requested):
Thanks, this is so much better!

I spotted one more small thing (unused variable), and then I think we're good to merge :)

Turro75:
Hi Uri, 
I added an automatic fix of mismatched registers so gdbdiff can continue to run endless reporting every error but keeping the emulator synced with silicone.

I've seen that most errors are due to ldr loading from missing peripherals registers, to make gdbdiff even more helpful it could report the asm at pc memory and print out which memory address stored a wrong value or printing out the code for a test ready to be pasted in the test file.

debug/gdbdiff.ts
```diff
@@ -50,10 +49,15 @@ function printFlags(xpsr: number) {
   return `[${negative}${zero}${carry}${overflow}]`;
 }
 
-function compareRegisters(emulator: Uint32Array, silicone: Uint32Array) {
+async function compareFixRegisters(
+  tofixclient: GDBClient,
```
urish, line 53: please follow the variable name convention, e.g. `toFixClient` (camel-case).
Also, it makes more sense to me to put this argument last (because the method name if compareFix, I expect the first arguments to be related to compare, and then the next ones to be compared to the fix). I Hope this makes sense

debug/gdbdiff.ts
```diff
@@ -69,23 +73,21 @@ async function main() {
...
-      printComparedRegisters(lastRegSet1, regSet1, regSet2);
-      process.exit(1);
+      printComparedRegisters(prevRegSet, regSet1, regSet2);
     }
-    lastRegSet1 = regSet1;
+    prevRegSet = regSet1;
     await client1.singleStep();
     await client2.singleStep();
-    if (counter % 500 === 0) {
+    if (counter % 50 === 0) {
```
urish, line 90: What's the reason for choosing 50?

debug/gdbdiff.ts
```diff
@@ -23,25 +67,29 @@ function compareRegisters(emulator: Uint32Array, silicone: Uint32Array) {
 async function main() {
   const client1 = new GDBClient();
   const client2 = new GDBClient();
-  await client1.connect('localhost');
-  await client2.connect('raspberrypi');
+  await client1.connect('localhost', 3334);
+  await client2.connect('localhost', 3333);
```
urish, line 71: I'm thinking, it may make more sense to call these `emulatorClient` and `siliconClient` instead of `client1`/`client2`. I know I chose those names, but now when I'm thinking of it, `client1` / `client2` are pretty bad names...

debug/gdbdiff.ts
```diff
@@ -23,25 +67,29 @@ function compareRegisters(emulator: Uint32Array, silicone: Uint32Array) {
...
   // Disable interrupts
   await client1.writeRegister(19, 1, 8);
   await client2.writeRegister(19, 1, 8);
   // Start diffing
-  for (let counter = 0; ; counter++) {
+  let prevRegSet = await client2.readRegisters();
+  let counter = 1;
```
urish, line 77: No need to separately delcare `counter`, as it is only used inside the `for` loop.

We can move the `let` to the for statement, e.g. `for (let counter = 1; ; counter++) {`

urish (changes requested):
Thanks Valerio! 

I like the idea of fixing the state and continuing to run. Added a few small comments/suggestions and a question.

Have you found anything new using the tool yet? 

I remember there was an issue with `main()`: if you load the test-printf program (from #24) and put a breakpoint in the beginning of `main()`, i.e. `break main`, you'll notice that r12 has a different value between the emulator and the silicone. So that's one thing we can look into, where is this different value coming from. Is it from some peripheral (like the timer)? Or some instruction set bug?

I think that at this point, if we still don't find anything new, this tool may be good enough for our purposes, and it may makes sense to shift the focus and look for other ways to find discrepancies between the simulator and the silicone. Thoughts?

urish:
About the `main()` issue, here are the hex/elf files for reproducing it:

[test-printf.zip](https://github.com/wokwi/rp2040js/files/6488903/test-printf.zip)

Running on silicone:

```
(gdb) monitor reset init
target halted due to debug-request, current mode: Thread 
xPSR: 0xf1000000 pc: 0x000000ee msp: 0x20041f00
target halted due to debug-request, current mode: Thread
xPSR: 0xf1000000 pc: 0x000000ee msp: 0x20041f00
(gdb) break main
Breakpoint 1 at 0x10000342: file /home/uri/p/test-printf/test-printf.ino, line 2.
(gdb) c
Continuing.
Note: automatically using hardware breakpoints for read-only addresses.
target halted due to debug-request, current mode: Thread 
xPSR: 0x01000000 pc: 0x00000178 msp: 0x20041f00

Thread 1 hit Breakpoint 1, main () at /home/uri/p/test-printf/test-printf.ino:2
2         printf("12 is %d\n", 12);
(gdb) info registers
r0             0x0                 0
r1             0x10001cd9          268442841
r2             0x20002858          536881240
r3             0x1                 1
r4             0x0                 0
r5             0x0                 0
r6             0x0                 0
r7             0x0                 0
r8             0x0                 0
r9             0x0                 0
r10            0x0                 0
r11            0x0                 0
r12            0x5                 5
sp             0x2000ceb0          0x2000ceb0 <_main_stack+32744>
lr             0x100040a7          268452007
pc             0x10000342          0x10000342 <main()+2>
xPSR           0x61000000          1627389952
msp            0x2003ffb8          0x2003ffb8
psp            0x2000ceb0          0x2000ceb0 <_main_stack+32744>
primask        0x0                 0
basepri        0x0                 0
faultmask      0x0                 0
control        0x2                 2
```

Running in the emulator:

```
0x10000000 in ?? ()
A program is being debugged already.
Are you sure you want to change the file? (y or n) y
Reading symbols from test-printf.ino.elf...
(gdb) break main
Breakpoint 1 at 0x10000342: file /home/uri/p/test-printf/test-printf.ino, line 2.
(gdb) c
Continuing.

Breakpoint 1, main () at /home/uri/p/test-printf/test-printf.ino:2
2         printf("12 is %d\n", 12);
(gdb) info registers
r0             0x0                 0
r1             0x10001cd9          268442841
r2             0x20002858          536881240
r3             0x1                 1
r4             0x0                 0
r5             0x0                 0
r6             0x0                 0
r7             0x0                 0
r8             0x0                 0
r9             0x0                 0
r10            0x0                 0
r11            0x0                 0
r12            0xa                 10
sp             0x2000ceb0          0x2000ceb0 <_main_stack+32744>
lr             0x100040a7          268452007
pc             0x10000342          0x10000342 <main()+2>
xPSR           0x61000000          1627389952
msp            0x2003ffb8          0x2003ffb8
psp            0x2000ceb0          0x2000ceb0 <_main_stack+32744>
primask        0x0                 0
basepri        0x0                 0
faultmask      0x0                 0
control        0x2                 2
(gdb)
```


difference:

Silicone - `r12            0x5                 5`
Emulator - `r12            0xa                 10`

urish:
Thanks! Anything else you want to change or can I merge this?

Turro75:
no I'm confident it is fine for now

urish:
Awesome, thanks!

urish:
Moved the information about the `main()` difference with r12 to a new issue: #28

Turro75:
[diff.txt](https://github.com/wokwi/rp2040js/files/6489015/diff.txt)
[testdiff.txt](https://github.com/wokwi/rp2040js/files/6489016/testdiff.txt)
This is the output of the emulator and gdbdiff of the hex You shared above, the funny thing is that mbed os crashed.
what I've seen are only errors due to unimplemented peripherals and mismatches from taken values. the asm seems working well for now. 

Now I'm trying to debug the pico prior to main to see where r12 is set to 5, using watchpoints on pico is so slow....

urish:
Thanks! 

The reason for Mbed OS crashing is probably the fact that we disabled the interrupts (writing 1 to PRIMASK). So it thinks we're inside an interrupt handler, and then if we try to call certain functions, it crashes.

One possible workaround would be to keep the interrupts enabled, but to write 0xffffffff to `NVIC_ICER` (at 0xe000e180) after each instruction, so all interrupts will remain disabled even though PRIMASK is 1.

Another workaround will probably be to find memory address of the instruction that reads from PRIMASK in Mbed OS code, and change the result so it always reads 0 (and then doesn't think we are inside interrupt handler)

urish:
And maybe just removing the code that writes 1 to PRIMASK will also work ?

urish:
Also, let's move this discussion to #28

PR #31, opened 2021-05-16
fix(rp2040): Fix CMP reg T1

Turro75:
found a specific test where CMPregT1 wrongly set the carry

### DIFF
--- a/src/instructions.spec.ts
+++ b/src/instructions.spec.ts
@@ -1055,19 +1055,6 @@ describe('Cortex-M0+ Instruction Set', () => {
     expect(registers.V).toEqual(false);
   });
 
-  it('should execute a `sbcs r0, r3` instruction', async () => {
-    await cpu.setPC(0x20000000);
-    await cpu.writeUint16(0x20000000, opcodeSBCS(r0, r3));
-    await cpu.setRegisters({ r0: 0, r3: 0xffffffff, C: false });
-    await cpu.singleStep();
-    const registers = await cpu.readRegisters();
-    expect(registers.r0).toEqual(0);
-    expect(registers.N).toEqual(false);
-    expect(registers.Z).toEqual(true);
-    expect(registers.C).toEqual(false);
-    expect(registers.V).toEqual(false);
-  });
-
   it('should execute a `sdmia r0!, {r1, r2}` instruction', async () => {
     await cpu.setPC(0x20000000);
     await cpu.writeUint16(0x20000000, opcodeSTMIA(r0, (1 << r1) | (1 << r2)));
--- a/src/instructions.spec.ts
+++ b/src/instructions.spec.ts
@@ -469,7 +469,7 @@ describe('Cortex-M0+ Instruction Set', () => {
     expect(registers.V).toEqual(false);
   });
 
-  it('should correctly unset carry when executing an `cmp	r2, r0` instruction', async () => {
+  it('should execute an cmp r2, r0 instruction and not set any flags when r0=0xb71b0000 and r2=0x00b71b00', async () => {
     await cpu.setPC(0x20000000);
     await cpu.writeUint16(0x20000000, opcodeCMPregT1(r2, r0));
     await cpu.setRegisters({ r0: 0xb71b0000, r2: 0x00b71b00 });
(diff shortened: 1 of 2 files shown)


src/instructions.spec.ts
```diff
@@ -469,6 +469,19 @@ describe('Cortex-M0+ Instruction Set', () => {
...
+    await cpu.setRegisters({ r0: 0xb71b0000 });
+    await cpu.setRegisters({ r2: 0x00b71b00 });
+    await cpu.setRegisters({ xPSR: 0x61000000 });
+    await cpu.singleStep();
+    const registers = await cpu.readRegisters();
+    expect(registers.N).toEqual(false);
+    expect(registers.Z).toEqual(false);
+    expect(registers.C).toEqual(false);
+    expect(registers.V).toEqual(false);
+  });
```
urish, line 484: can you please add an empty line after the end of the test case?

Turro75:
Please be patient, one Day I'll learn how to PR only the relevant commits...
0cdd227 is the right one

src/instructions.spec.ts
```diff
@@ -469,6 +469,19 @@ describe('Cortex-M0+ Instruction Set', () => {
     expect(registers.V).toEqual(false);
   });
 
+  it('should execute an `cmp	r2, r0` instruction', async () => {
+    await cpu.setPC(0x20000000);
+    await cpu.writeUint16(0x20000000, opcodeCMPregT1(r2, r0));
+    await cpu.setRegisters({ r0: 0xb71b0000 });
+    await cpu.setRegisters({ r2: 0x00b71b00 });
+    await cpu.setRegisters({ xPSR: 0x61000000 });
```
urish, line 477: Instead of setting xPSR directly, let's set the specific flags (like we did in the "should execute `add sp, r8` instruction and not update the flags" test case)

urish: Also, we can have a single call to setRegisters setting all the registers at once (instead of one call per register)

src/instructions.spec.ts
```diff
@@ -469,6 +469,19 @@ describe('Cortex-M0+ Instruction Set', () => {
     expect(registers.V).toEqual(false);
   });
 
+  it('should execute an `cmp	r2, r0` instruction', async () => {
```
urish, line 472: Let's make the test name more specific, like: "should execute an `cmp	r2, r0` instruction and not set any flags when r0=0xb71b0000 and r2=0x00b71b00"

urish (changes requested):
Thanks for finding all these small flag bugs in the emulator :)

I've added some comments!

urish:
> Please be patient, one Day I'll learn how to PR only the relevant commits...

if you need any help with that let me know

src/instructions.spec.ts
```diff
@@ -469,19 +469,18 @@ describe('Cortex-M0+ Instruction Set', () => {
...
   });
 
-  it('should execute an `cmp	r2, r0` instruction', async () => {
+  it('should correctly unset carry when executing an `cmp	r2, r0` instruction', async () => {
     await cpu.setPC(0x20000000);
     await cpu.writeUint16(0x20000000, opcodeCMPregT1(r2, r0));
-    await cpu.setRegisters({ r0: 0xb71b0000 });
-    await cpu.setRegisters({ r2: 0x00b71b00 });
-    await cpu.setRegisters({ xPSR: 0x61000000 });
+    await cpu.setRegisters({ r0: 0xb71b0000, r2: 0x00b71b00 });
```
urish, line 475: Didn't you also want to set C/Z to true?

Turro75: I tested it now, the previous flag can be any value so I removed the set.

urish: okay great!

src/instructions.spec.ts
```diff
@@ -469,6 +469,18 @@ describe('Cortex-M0+ Instruction Set', () => {
     expect(registers.V).toEqual(false);
   });
 
+  it('should execute an cmp r2, r0 instruction and not set any flags when r0=0xb71b0000 and r2=0x00b71b00', async () => {
```
urish, line 472: Ok, one more small thing: please add backticks (\`) around the instruction name, like in the other tests

PR #37, opened 2021-05-21
feat(sio): fix implemented Hardware Divider

Turro75:
thanks @Kilograham for pointing me in the right direction

### DIFF
--- a/src/sio.spec.ts
+++ b/src/sio.spec.ts
@@ -1,64 +1,82 @@
 import { RP2040, SIO_START_ADDRESS } from './rp2040';
 
+import { GDBClient } from './utils/gdbclient';
+import { ICortexTestDriver } from './utils/test-driver';
+import { GDBTestDriver } from './utils/test-driver-gdb';
+import { RP2040TestDriver } from './utils/test-driver-rp2040';
+
 //HARDWARE DIVIDER
-const _DIV_UDIVIDEND = SIO_START_ADDRESS + 0x060; //  Divider unsigned dividend
-const _DIV_UDIVISOR = SIO_START_ADDRESS + 0x064; //  Divider unsigned divisor
-const _DIV_SDIVIDEND = SIO_START_ADDRESS + 0x068; //  Divider signed dividend
-const _DIV_SDIVISOR = SIO_START_ADDRESS + 0x06c; //  Divider signed divisor
-const _DIV_QUOTIENT = SIO_START_ADDRESS + 0x070; //  Divider result quotient
-const _DIV_REMAINDER = SIO_START_ADDRESS + 0x074; //Divider result remainder
-const _DIV_CSR = SIO_START_ADDRESS + 0x078;
+const SIO_DIV_UDIVIDEND = SIO_START_ADDRESS + 0x060; //  Divider unsigned dividend
+const SIO_DIV_UDIVISOR = SIO_START_ADDRESS + 0x064; //  Divider unsigned divisor
+const SIO_DIV_SDIVIDEND = SIO_START_ADDRESS + 0x068; //  Divider signed dividend
+const SIO_DIV_SDIVISOR = SIO_START_ADDRESS + 0x06c; //  Divider signed divisor
+const SIO_DIV_QUOTIENT = SIO_START_ADDRESS + 0x070; //  Divider result quotient
+const SIO_DIV_REMAINDER = SIO_START_ADDRESS + 0x074; //Divider result remainder
+const SIO_DIV_CSR = SIO_START_ADDRESS + 0x078;
 
 describe('RPSIO', () => {
+  let cpu: ICortexTestDriver;
+
+  beforeEach(async () => {
+    // if (process.env.TEST_GDB_SERVER) {
+    if (false) {
+      const client = new GDBClient();
+      await client.connect('127.0.0.1');
+      cpu = new GDBTestDriver(client);
+      await cpu.init();
+    } else {
+      cpu = new RP2040TestDriver(new RP2040());
+    }
+  });
+
+  afterEach(async () => {
+    await cpu.tearDown();
+  });
+
   describe('Hardware Divider', () => {
-    it('should set perform a signed hardware divider 123456 / -321 = -384 R192', () => {
-      const rp2040 = new RP2040();
-      expect(rp2040.readUint32(_DIV_CSR)).toEqual(0);
-      rp2040.writeUint32(_DIV_SDIVIDEND, 123456);
-      rp2040.writeUint32(_DIV_SDIVISOR, -321);
-      expect(rp2040.readUint32(_DIV_CSR)).toEqual(1);
-      expect(rp2040.readUint32(_DIV_REMAINDER)).toEqual(192);
-      expect(rp2040.readUint32(_DIV_QUOTIENT)).toEqual(-384);
-      expect(rp2040.readUint32(_DIV_CSR)).toEqual(0);
+    it('should set perform a signed hardware divider 123456 / -321 = -384 R192', async () => {
+      await cpu.writeUint32(SIO_DIV_SDIVIDEND, 123456);
+      expect(await cpu.readUint32(SIO_DIV_SDIVIDEND)).toEqual(123456);
+      await cpu.writeUint32(SIO_DIV_SDIVISOR, -321);
+      expect(await cpu.readUint32(SIO_DIV_CSR)).toEqual(3);
+      expect((await cpu.readUint32(SIO_DIV_SDIVISOR)) | 0).toEqual(-321);
+      expect((await cpu.readUint32(SIO_DIV_REMAINDER)) | 0).toEqual(192);
+      expect((await cpu.readUint32(SIO_DIV_QUOTIENT)) | 0).toEqual(-384);
+      expect(await cpu.readUint32(SIO_DIV_CSR)).toEqual(1);
     });
 
-    it('should set perform an unsigned hardware divider 123456 / 321 = 384 R192', () => {
-      const rp2040 = new RP2040();
-      expect(rp2040.readUint32(_DIV_CSR)).toEqual(0);
-      rp2040.writeUint32(_DIV_UDIVIDEND, 123456);
-      const cycles = rp2040.cycles;
-      rp2040.writeUint32(_DIV_UDIVISOR, 321);
-      expect(rp2040.cycles - cycles).toEqual(8);
-      expect(rp2040.readUint32(_DIV_CSR)).toEqual(1);
-      expect(rp2040.readUint32(_DIV_REMAINDER)).toEqual(192);
-      expect(rp2040.readUint32(_DIV_QUOTIENT)).toEqual(384);
-      expect(rp2040.readUint32(_DIV_CSR)).toEqual(0);
+    it('should set perform an unsigned hardware divider 123456 / 321 = 384 R192', async () => {
+      await cpu.writeUint32(SIO_DIV_UDIVIDEND, 123456);
+      await cpu.writeUint32(SIO_DIV_UDIVISOR, 321);
+      expect(await cpu.readUint32(SIO_DIV_CSR)).toEqual(3);
+      expect(await cpu.readUint32(SIO_DIV_REMAINDER)).toEqual(192);
+      expect(await cpu.readUint32(SIO_DIV_QUOTIENT)).toEqual(384);
+      expect(await cpu.readUint32(SIO_DIV_CSR)).toEqual(1);
     });
 
-    it('should perform a division, store the result, do another division then restore the previously stored result ', () => {
-      const rp2040 = new RP2040();
-      rp2040.writeUint32(_DIV_SDIVIDEND, 123456);
-      rp2040.writeUint32(_DIV_SDIVISOR, -321);
-      const remainder = rp2040.readUint32(_DIV_REMAINDER);
-      const quotient = rp2040.readUint32(_DIV_QUOTIENT);
+    it('should perform a division, store the result, do another division then restore the previously stored result ', async () => {
+      await cpu.writeUint32(SIO_DIV_SDIVIDEND, 123456);
+      await cpu.writeUint32(SIO_DIV_SDIVISOR, -321);
+      const remainder = (await cpu.readUint32(SIO_DIV_REMAINDER)) | 0;
+      const quotient = (await cpu.readUint32(SIO_DIV_QUOTIENT)) | 0;
       expect(remainder).toEqual(192);
       expect(quotient).toEqual(-384);
-      rp2040.writeUint32(_DIV_UDIVIDEND, 123);
-      rp2040.writeUint32(_DIV_UDIVISOR, 7);
-      expect(rp2040.readUint32(_DIV_REMAINDER)).toEqual(4);
-      expect(rp2040.readUint32(_DIV_QUOTIENT)).toEqual(17);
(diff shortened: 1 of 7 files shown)


urish:
Thanks Valerio! 

Yes, Graham surely knows the RP2040 much better than me. I'm so happy he's helping us :-)

Do you want us to add some unit tests for the divider in the next live stream? Or do you want to try adding them yourself?

src/sio.spec.ts
```diff
@@ -0,0 +1,64 @@
+import { RP2040, SIO_START_ADDRESS } from './rp2040';
+
+//HARDWARE DIVIDER
+const _DIV_UDIVIDEND = SIO_START_ADDRESS + 0x060; //  Divider unsigned dividend
+const _DIV_UDIVISOR = SIO_START_ADDRESS + 0x064; //  Divider unsigned divisor
+const _DIV_SDIVIDEND = SIO_START_ADDRESS + 0x068; //  Divider signed dividend
+const _DIV_SDIVISOR = SIO_START_ADDRESS + 0x06c; //  Divider signed divisor
+const _DIV_QUOTIENT = SIO_START_ADDRESS + 0x070; //  Divider result quotient
+const _DIV_REMAINDER = SIO_START_ADDRESS + 0x074; //Divider result remainder
+const _DIV_CSR = SIO_START_ADDRESS + 0x078;
```
urish, line 10: Can you please remove the _ (underscore) in front of the constant? The idea is to keep this code consistent with the rest of the code

Turro75: ok, I'm moving test to be gdbclient ready, I'll do it in the next commit.

urish: Oh great. I was going to suggest that, but you are faster!

Turro75: ok done, 
I had to add several "| 0"which are needed when running on silicone and the read value is signed, maybe a readSInt32 implementation could be better? or maybe not as it matters only during test? 

copied the silicone behavior when divisor=0 -> quot=-1, rem=dividend
I wasn't able to find any word related to div/0 exceptions in the datasheets, is this possible?

_DIV_XXX has been renamed to SIO_DIV_XXX, I choose this way as SIO_DIV_XXX = SIO_ADDRESS_BASE + DIV_XXX
Even if declared in different files, I wouldn't have the same name to both offset and full address.

src/sio.ts
```diff
@@ -117,6 +146,56 @@ export class RPSIO {
       case GPIO_HI_OE_XOR:
         this.qspiGpioOutputEnable ^= value & GPIO_MASK;
         break;
+      case DIV_UDIVIDEND:
+        this.divDividend = value >>> 0;
+        this.divCSR = 0b10;
```
urish, line 151: what's the reason for updating divCSR twice in the same switch case?

src/sio.ts
```diff
@@ -117,6 +146,56 @@ export class RPSIO {
...
+          this.divQuotient = Math.trunc(this.divDividend / this.divDivisor) >>> 0;
+          this.divRemainder = Math.trunc(this.divDividend % this.divDivisor) >>> 0;
+        }
+        this.divCSR = 0b11;
+        this.rp2040.cycles += 8;
+        break;
+      case DIV_SDIVISOR:
+        this.divCSR = 0b10;
+        this.divDivisor = value | 0;
+        if (this.divDivisor == 0) {
```
urish, line 181: I suggest that we move this repetitive code (in DIV_UDIVIDEND, DIV_SDIVIDEND, DIV_UDIVISOR, and DIV_SDIVISOR) into a function that will do the actual division, e.g. `this.updateDivider()`.

As far as I can tell, you don't need to worry about the sign of the result, because when it's read back into a register it'll always be internally converted to unsigned anyway.

src/sio.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+      cpu = new RP2040TestDriver(new RP2040());
+    }
+  });
+
+  afterEach(async () => {
+    await cpu.tearDown();
+  });
+
+  describe('Hardware Divider', () => {
+    it('should set perform a signed hardware divider 123456 / -321 = -384 R192', async () => {
```
urish, line 37: I think "REM 192" will be clearer than "R192" (same for the other test names).

Also, the word "set" (second word) seems not related here?

src/sio.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+      expect(await cpu.readUint32(SIO_DIV_REMAINDER)).toEqual(4);
+      expect(await cpu.readUint32(SIO_DIV_QUOTIENT)).toEqual(17);
+      await cpu.writeUint32(SIO_DIV_REMAINDER, remainder);
+      await cpu.writeUint32(SIO_DIV_QUOTIENT, quotient);
+      expect(await cpu.readUint32(SIO_DIV_CSR)).toEqual(3);
+      expect((await cpu.readUint32(SIO_DIV_REMAINDER)) | 0).toEqual(192);
+      expect((await cpu.readUint32(SIO_DIV_QUOTIENT)) | 0).toEqual(-384);
+    });
+
+    it('should set perform an unsigned division by zero 123456 / 0 = Infinity RNaN', async () => {
```
urish, line 75: The result isn't Infinity / NaN (as the test name says), rather 0xffffffff and remainder 123456.

urish (changes requested):

src/sio.spec.ts
urish: Thank you!

> copied the silicone behavior when divisor=0 -> quot=-1, rem=dividend
Good idea to test what happens when dividing by zero. 

> I wasn't able to find any word related to div/0 exceptions in the datasheets, is this possible?

No idea! Did you see it getting any exception on the silicone? 

> I had to add several "| 0"which are needed when running on silicone and the read value is signed, maybe a readSInt32 implementation could be better? or maybe not as it matters only during test?

I think this is a good idea, it will make the tests easier to read / understand. I suggest calling it `readInt32()` (without the S), similar to how they call the [signed int array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Int32Array) in JavaScript (and also C++ type [int32_t vs uint32_t](https://en.cppreference.com/w/cpp/types/integer)).

> _DIV_XXX has been renamed to SIO_DIV_XXX, I choose this way as SIO_DIV_XXX = SIO_ADDRESS_BASE + DIV_XXX
Even if declared in different files, I wouldn't have the same name to both offset and full address.

👍

src/sio.ts
Turro75: I was waiting for You on this...
well the datasheet states that the CSR lsb is cleared at the beginning of the calculation and set once result is ready (after 8 cpu cycles).
On a single core this assignment does not make any sense, since I don't know how multicore will be implemented it could be something that will matter in case of concurrency against hardware divider.

urish: That's a good point. When we get to multicore, we'll just execute one instruction at a time (so jumping back and forth between running core0 and core1 code). 

In any case, JavaScript does not support multi-threading natively. The only way is using Web Workers (or [worker threads](https://nodejs.org/api/worker_threads.html) on Node.js). But even then, the only way two threads can access the same memory is using a SharedArrayBuffer, so if we decide at some point to truly run this on two threads (and share the divider memory between them) we'll have to change all these variables to be stored inside a SharedArrayBuffer and refactor the the code, and I don't see it happening any time soon.

So to summarize, I think we can safely remove the `this.divCSR = 0b10;`

src/sio.spec.ts
```diff
@@ -34,18 +33,18 @@ describe('RPSIO', () => {
...
-      expect(await cpu.readUint32(SIO_DIV_SDIVIDEND)).toEqual(123456);
+      expect(await cpu.readInt32(SIO_DIV_SDIVIDEND)).toEqual(123456);
       await cpu.writeUint32(SIO_DIV_SDIVISOR, -321);
       expect(await cpu.readUint32(SIO_DIV_CSR)).toEqual(3);
-      expect((await cpu.readUint32(SIO_DIV_SDIVISOR)) | 0).toEqual(-321);
-      expect((await cpu.readUint32(SIO_DIV_REMAINDER)) | 0).toEqual(192);
-      expect((await cpu.readUint32(SIO_DIV_QUOTIENT)) | 0).toEqual(-384);
+      expect(await cpu.readInt32(SIO_DIV_SDIVISOR)).toEqual(-321);
+      expect(await cpu.readInt32(SIO_DIV_REMAINDER)).toEqual(192);
+      expect(await cpu.readInt32(SIO_DIV_QUOTIENT)).toEqual(-384);
```
urish, line 43: So much cleaner :)

src/utils/test-driver-gdb.ts
```diff
@@ -148,4 +148,9 @@ export class GDBTestDriver implements ICortexTestDriver {
     const result = await this.gdbClient.readMemory(address, 4);
     return new Uint32Array(result.buffer)[0];
   }
+
+  async readInt32(address: number) {
+    const result = await this.gdbClient.readMemory(address, 4);
+    return new Uint32Array(result.buffer)[0] | 0;
```
urish, line 154: why not `this.readUint32(address) | 0` like you did in the RP2040 test driver?

src/sio.ts
```diff
@@ -147,46 +159,14 @@ export class RPSIO {
         this.qspiGpioOutputEnable ^= value & GPIO_MASK;
         break;
       case DIV_UDIVIDEND:
-        this.divDividend = value >>> 0;
```
urish: Does it work correctly even though you use the same `this.divDividend = value;` for both the signed/unsigned cases (instead of `value >>> 0` vs `value | 0`)?

src/sio.spec.ts
```diff
@@ -1,11 +1,10 @@
 import { RP2040, SIO_START_ADDRESS } from './rp2040';
-
 import { GDBClient } from './utils/gdbclient';
 import { ICortexTestDriver } from './utils/test-driver';
 import { GDBTestDriver } from './utils/test-driver-gdb';
 import { RP2040TestDriver } from './utils/test-driver-rp2040';
 
-//HARDWARE DIVIDER
+//HARDWARE DIVIDER REGISTERS ABSOLUTE ADDRESS
```
urish, line 7: Can you please change to 

`// Hardware Divider registers absolute address`

(so not all letters are capital)

src/sio.ts
```diff
@@ -46,6 +46,18 @@ export class RPSIO {
 
   constructor(private readonly rp2040: RP2040) {}
 
+  updateHardwareDivider() {
+    if (this.divDivisor == 0) {
+      this.divQuotient = 0xffffffff;
+      this.divRemainder = this.divDividend;
```
urish, line 52: Shouldn't we update divCSR / cycles in this case as well?

urish (changes requested):
Thanks Valerio! I added a few comments / suggestions. I like the shape this is taking :)

src/sio.ts
Turro75: in the end yes, it works in both cases.
I don't understand the reason to have separated registers, at the beginning I though putting a boolean flag to change the behavior between signed/unsigned, it turned out a useless handling.

urish: Have you tried dividing -3000 by 2?

In JavaScript I get different results (printing both results as signed, just to show the difference):

```js
const signedDiv = (a, b) => ((a|0)/(b|0))|0;
const unsignedDiv = (a, b) => ((a>>>0)/(b>>>0))|0;

console.log(signedDiv(-3000, 2)); // -1500
console.log(unsignedDiv(-3000, 2)); // 2147482148
```

src/sio.ts
Turro75: correct, sorry stupid bug.

src/sio.ts
Turro75: I added this test, it passes on both emu and sil.
what I've seen is that in the emu it doesn't matter if I use readUint32 or readInt32 it always return the right value.
test in the silicon instead it changes and it requires the right read method.
Seems that javascript (math.trunc?) automatically handles the right way in case of negative value.

urish: For the result, it shouldn't matter how you read it (you can convert between signed / unsigned at any stage).

But for the actual division, it seems to matter even if you use math.trunc:

```js
const signedDiv = (a, b) => Math.trunc((a|0)/(b|0))|0;
const unsignedDiv = (a, b) => Math.trunc((a>>>0)/(b>>>0))|0;

console.log(signedDiv(-3000, 2)); // -1500
console.log(unsignedDiv(-3000, 2)); // 2147482148
```

It could be the the current code is always doing signed division, and that's why the test passes. In this case, you can add a test for unsigned divide as well, trying to divide 0x80000000 by 2 using unsigned integer divide, and verify that you get the correct result (0x40000000):

```js
console.log((signedDiv(0x80000000, 2) >>> 0).toString(16)); // 0xc0000000
console.log((unsignedDiv(0x80000000, 2) >>> 0).toString(16)); // 0x40000000 
```

src/sio.ts
```diff
@@ -24,14 +24,45 @@ const GPIO_HI_OE_XOR = 0x04c; // QSPI output enable XOR
...
+  divDivisor = 1;
+  divQuotient = 0;
+  divRemainder = 0;
+  divCSR = 0;
 
   constructor(private readonly rp2040: RP2040) {}
 
+  updateHardwareDivider(signed: boolean) {
+    if (this.divDivisor == 0) {
+      this.divQuotient = (this.divDividend | 0) < 0 ? 1 : 0xffffffff;
```
urish, line 51: Is this true both for signed / unsigned divide?

src/sio.ts
kilograham: note that SIO dividers/interpreters are core local

urish: That's a good point to remember when we implemented multi-core. I noted this!

src/sio.ts
Turro75: arghhhh!!!  missing break forced signed code.
I've also had to force the readUint32 to return unsigned to pass this test:
`should perform a signed division 0x80000000 / 2 = 0xc0000000 REM 0`

src/sio.ts
Turro75: this is correct (and works) for all 3 conditions
this.divQuotient = this.divDividend > 0 ? -1 : 1;

urish:
Thanks Valerio, this looks good to me :-)

Is there anything else you want to add/change before I merge?

Turro75:
Oh no  Uri, I think this divider took enough time. It is fine now.

urish:
Alright, merged! 

Thanks for working on it, it was a long one indeed... But now I feel the instruction set is truly complete (this hardware divider is like another CPU instruction). We'll review the implementation in the live stream tomorrow :-)

Turro75:
Hi Uri,

I want to say sorry for forcing You to stay so much time on this, I'm sure You would have successfully write it in less than 20 minutes.
I wasn't really aware of signed/unsigned trap of javascript (after many years on freepascal) and forgot to set relevant breaks in switch/case. My regrets are that mostly I sent many commits too early.
Now I'm PR spinlocks which instead I'm pretty sure the code is ready to be merged.

urish:
Hi Valerio,

First of all, thanks for caring about my time! I'm very pleased with the results. Maybe just writing the code would take me 20 minutes, but as you've seen in the live streams, writing the test cases, finding the bugs + verifying with the silicone would probably take much more. Like the alarms last week took about 2 hours.

Anyway, all the bugs you've found so far with gdbdiff (and fixed) surely saved much precious time. And I'm happy to see your commits and provide feedback. This way everyone benefits :)

PR #38, opened 2021-05-22
feat(gdbdiff): use the bootrom code from real pico

Turro75:
github published bootrom may differs from bootrom of a real pico.  
dumping (within gdb session)  the bootrom hex from real pico allows a better use of gdbdiff when code jumps to bootrom (i.e. by using hardware divider example)

### DIFF
--- a/demo/emulator-run.ts
+++ b/demo/emulator-run.ts
@@ -1,14 +1,21 @@
 import * as fs from 'fs';
 import { RP2040 } from '../src';
-import { bootromB1 } from './bootrom';
+//import { bootromB1 } from './bootrom';
 import { loadHex } from './intelhex';
 import { GDBTCPServer } from '../src';
 
 // Create an array with the compiled code of blink
 // Execute the instructions from this array, one by one.
 const hex = fs.readFileSync('hello_uart.hex', 'utf-8');
+//bootromhex taken from pico by "dump ihex memory bootrom.hex 0x0 0x3fff"
+const bootromhex = fs.readFileSync('demo/bootrom.hex', 'utf-8');
 const mcu = new RP2040();
-mcu.loadBootrom(bootromB1);
+let localbootrom = new ArrayBuffer(0x4000);
+let localbootromview8 = new Uint8Array(localbootrom);
+let localbootromview32 = new Uint32Array(localbootrom);
+loadHex(bootromhex, localbootromview8, 0x00000000);
+mcu.loadBootrom(localbootromview32);
+//mcu.loadBootrom(bootromB1);
 loadHex(hex, mcu.flash, 0x10000000);
 
 const gdbServer = new GDBTCPServer(mcu, 3333);
@@ -19,4 +26,4 @@ mcu.uart[0].onByte = (value) => {
 };
 
 mcu.PC = 0x10000000;
-mcu.execute();
+//mcu.execute();
--- a/demo/emulator-run.ts
+++ b/demo/emulator-run.ts
@@ -1,24 +1,33 @@
 import * as fs from 'fs';
 import { RP2040 } from '../src';
-//import { bootromB1 } from './bootrom';
+import { bootromB1 } from './bootrom';
 import { loadHex } from './intelhex';
 import { GDBTCPServer } from '../src';
 
 // Create an array with the compiled code of blink
 // Execute the instructions from this array, one by one.
-const hex = fs.readFileSync('hello_uart.hex', 'utf-8');
-//bootromhex taken from pico by "dump ihex memory bootrom.hex 0x0 0x3fff"
-const bootromhex = fs.readFileSync('demo/bootrom.hex', 'utf-8');
+const hex = fs.readFileSync(
+  '/home/turro/dev/arduino/raspberrypipico/blink2/blink2/Build/blink2.ino.hex',
+  'utf-8'
+);
 const mcu = new RP2040();
-let localbootrom = new ArrayBuffer(0x4000);
-let localbootromview8 = new Uint8Array(localbootrom);
-let localbootromview32 = new Uint32Array(localbootrom);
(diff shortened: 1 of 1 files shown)


urish:
Thanks for creating a separate branch for this!

It's interesting, maybe the reason for the difference is that the Pi Pico has the older version of the bootrom, V0?

In any case, it may make sense to either:
1. Include the dumped bootrom in the demo directory
- or - 
2. Make it possible to choose whether to load the compiled or dumped bootrom

so people without the physical chip / access to GDB will still be able to run the demo project. Thoughts?

Turro75:
It requires its time but even Valerio is able to learn something :-)

the reason to dump the pico memory it is only to allow gdbdiff running against a real chip straight even when the code uses bootrom sections.
If gdbdiff or testdriver aren't used or the user doesn't own a real pico there are no reason to use the dumped against the compiled.
I tried compiling v0 v1 with and without symbols, all differ from dumped code, may be a compiler setup mismatch. 

I didn't include the bootrom.hex because I'm not sure every pico in the market has the same, I think it is better let every chip owner who is interested in doing debugging stuff dumping it's own.
I could modify the code : if exists a bootrom.hex load it otherwise use the compiled bootrom.ts. Do You agree?

urish:
> I didn't include the bootrom.hex because I'm not sure every pico in the market has the same, I think it is better let every chip owner who is interested in doing debugging stuff dumping it's own.
> I could modify the code : if exists a bootrom.hex load it otherwise use the compiled bootrom.ts. Do You agree?

That's a good point. How about just loading the bootrom from the chip using the GDBClient? 
This way the user won't even have to worry about dumping it and all the stuff, it'll be done automatically. I'm thinking along the lines of:

1. Add a new method to dump a chunk of memory into a buffer, e.g. `readMemory(offset: number, count: number)`
2. Add a new method to write a buffer into memory, e.g. `writeMemory(offset: number, buffer: Uint32Array)`
3. Then do something like: `await emulatorClient.writeMemory(0, await siliconeClient.writeMemory(0, 0x4000));`

We'll also need to change gdbserver to make it possible to write to the bootrom. One way to do it would just be to use `rp2040.bootrom.set(...)` to update the bootrom in place. Another way would just be to add an `offset` parameter to `setBootrom()` (with a default value of 0), so it would be useful for partial writes to the bootrom.

Thoughts?

Turro75:
> That's a good point. How about just loading the bootrom from the chip using the GDBClient?
> This way the user won't even have to worry about dumping it and all the stuff, it'll be done automatically. I'm thinking along the lines of:
> 
> 1. Add a new method to dump a chunk of memory into a buffer, e.g. `readMemory(offset: number, count: number)`
> 2. Add a new method to write a buffer into memory, e.g. `writeMemory(offset: number, buffer: Uint32Array)`
> 3. Then do something like: `await emulatorClient.writeMemory(0, await siliconeClient.writeMemory(0, 0x4000));`
> 
> We'll also need to change gdbserver to make it possible to write to the bootrom. One way to do it would just be to use `rp2040.bootrom.set(...)` to update the bootrom in place. Another way would just be to add an `offset` parameter to `setBootrom()` (with a default value of 0), so it would be useful for partial writes to the bootrom.
> 
> Thoughts?

I think this will take ages to me, once I finished You already released the arduino nano rp2040 emulator...
in the meanwhile I simply add a control that check if bootrom.hex exists then load it, otherwise load the compiled bootrom.

did You mean this?
 3. Then do something like: `await emulatorClient.writeMemory(0, await siliconeClient.**read**Memory(0, 0x4000));`

urish:
> I think this will take ages to me, once I finished You already released the arduino nano rp2040 emulator...
in the meanwhile I simply add a control that check if bootrom.hex exists then load it, otherwise load the compiled bootrom.

We can also do it together on the live stream if you prefer. 

> did You mean this?
> 3. Then do something like: `await emulatorClient.writeMemory(0, await siliconeClient.**read**Memory(0, 0x4000));`

Yes, exactly!

Turro75:
> We can also do it together on the live stream if you prefer.

ok together with automatic alignment of the gdbdiff starting point. it will deserve a dedicated stream.

urish:
> automatic alignment of the gdbdiff starting point. 

What does it mean?

Turro75:
It means now the user has to set the starting point on both emulator and silicone then start gdbdiff.
trying to automate this would be smart too?

urish:
> trying to automate this would be smart too?

Oh, I get it. Well, how do you suggest to do that?

urish:
By the way, I found a way to disable interrupts (on silicone) without setting PRIMASK.PM to 1:

```
monitor cortex_m maskisr on
```

Turro75:
That's interesting, I always thought that implementing monitor on both gdbclient and gdbserver will be an improvement. Do You think this has priority over peripherals to implementing them better?

urish:
We already have monitor in gdbclient :-)

With gdbserver, the thing is `monitor` isn't a standard things, the way it works depends on the debugger you are using. In case of OpenOCD, it just runs OpenOCD commands. I learnt about all the commands by running `monitor help`. 

If I understand correctly, OpenOCD is actually a [TCL](https://en.wikipedia.org/wiki/Tcl) interpreter, so you can use any TCL command, e.g. `monitor expr 5 * 50 + [llength [list "a" "b" "c"]]`. So implementing full OpenOCD compatibility will be a lot of work. But we can probably implement just a few commands useful for us, e.g. the maskisr command I mention above, and maybe also "reset" (so you can restart the simulation without restarting the node.js process). Just some random thoughts :)

kilograham:
Note if you use the compilers mentioned in the README.md for branch b0/b1 in the bootrom repo then you should get an identical binary to what is in the chips (save for a 4 byte git revision which is likely not ever read by the code you are running).

B0 is likely only found on Picos. All other boards and future Picos are B1.

PR #40, opened 2021-05-24
feat(sio):implemented Spinlocks

Turro75:
I choose to avoid 32 spinlocks cases by inserting  if statements in both read/writeUint32

### DIFF
--- a/src/sio.ts
+++ b/src/sio.ts
@@ -70,152 +70,144 @@ export class RPSIO {
   }
 
   readUint32(offset: number) {
-    if (offset < SPINLOCK0) {
-      //not a spinlock
-      switch (offset) {
-        case GPIO_IN:
-          return 0; // TODO implement!
-        case GPIO_HI_IN:
-          return 0; // TODO implement!
-        case GPIO_OUT:
-          return this.gpioValue;
-        case GPIO_OE:
-          return this.gpioOutputEnable;
-        case GPIO_HI_OUT:
-          return this.qspiGpioValue;
-        case GPIO_HI_OE:
-          return this.qspiGpioOutputEnable;
-        case GPIO_OUT_SET:
-        case GPIO_OUT_CLR:
-        case GPIO_OUT_XOR:
-        case GPIO_OE_SET:
-        case GPIO_OE_CLR:
-        case GPIO_OE_XOR:
-        case GPIO_HI_OUT_SET:
-        case GPIO_HI_OUT_CLR:
-        case GPIO_HI_OUT_XOR:
-        case GPIO_HI_OE_SET:
-        case GPIO_HI_OE_CLR:
-        case GPIO_HI_OE_XOR:
-          return 0; // TODO verify with silicone
-        case CPUID:
-          // Returns the current CPU core id (always 0 for now)
-          return 0;
-        case SPINLOCK_ST:
-          return this.spinLock;
-        case DIV_UDIVIDEND:
-          return this.divDividend;
-        case DIV_SDIVIDEND:
-          return this.divDividend;
-        case DIV_UDIVISOR:
-          return this.divDivisor;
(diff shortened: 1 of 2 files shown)


Turro75:
I didn't find a way to set a case range to avoid 32 cases, any better than if statement?

src/sio.ts
```diff
@@ -64,129 +70,152 @@ export class RPSIO {
...
+          return this.divQuotient;
+        case DIV_REMAINDER:
+          return this.divRemainder;
+        case DIV_CSR:
+          return this.divCSR;
+      }
+    }
+    if (offset >= SPINLOCK0 && offset <= SPINLOCK31) {
+      //spinlock
+      const bitIndex = (1 << ((offset - SPINLOCK0) / 4)) >>> 0;
```
urish, line 125: no need for unsigned conversion (the bitwise | operation below will make it signed anyway...)

also, I think `bitMask` is a bit better name (because it's not the index of the bit, it's a value where that bit is set to 1)

src/sio.ts
```diff
@@ -64,129 +70,152 @@ export class RPSIO {
...
+        case DIV_QUOTIENT:
+          this.divCSR &= ~0b10;
+          return this.divQuotient;
+        case DIV_REMAINDER:
+          return this.divRemainder;
+        case DIV_CSR:
+          return this.divCSR;
+      }
+    }
+    if (offset >= SPINLOCK0 && offset <= SPINLOCK31) {
```
urish, line 123: I suggest to move this `if` above the `switch`, and then keep the switch as it is. Since this if always returns something, we'll only get to the switch if the offset is not one of the spinlocks...

e.g., this is the structure that I have in mind

```
if (offset >= SPINLOCK0 && offset <= SPINLOCK31) {
   // ...
}

switch (offset) {
   // ...
}

console.warn(`Read from invalid SIO address: ${offset.toString(16)}`);
return 0xffffffff;
```

src/sio.ts
```diff
@@ -64,129 +70,152 @@ export class RPSIO {
...
+          this.divQuotient = value;
+          this.divCSR = 0b11;
+          break;
+        case DIV_REMAINDER:
+          this.divRemainder = value;
+          this.divCSR = 0b11;
+          break;
+      }
+    }
+    if (offset >= SPINLOCK0 && offset <= SPINLOCK31) {
```
urish, line 215: same comment as above - moving the `if` above the `switch` and adding a `return` so it never gets to the switch in case of a spinlock.

src/sio.ts
urish: or `bitIndexMask` that you used below is also a good name

urish (changes requested):
Thanks! Left some comments about the structure of the code

urish:
And a quick note about ranges for switch in JavaScript: it is possible, but in a weird way. For example:

```javascript
switch (true) {
   case i === 0: return "zero";
   case i > 0 && i < 10: return "single digit";
   case i < 100: return "two digits";
   default: return "big number!";
}
```

But I don't think it's a good practice. Using an `if` statement is probably a better solution, especially in our case.

p.s. there's also a [proposal for new JavaScript syntax](https://github.com/tc39/proposal-pattern-matching) that will probably able to deal with ranges. But it's just an early stage proposal right now...

src/sio.ts
```diff
@@ -114,6 +131,10 @@ export class RPSIO {
   }
 
   writeUint32(offset: number, value: number) {
+    if (offset >= SPINLOCK0 && offset <= SPINLOCK31) {
+      const bitIndexMask = ~(1 << ((offset - SPINLOCK0) / 4));
+      this.spinLock &= bitIndexMask;
```
urish, line 136: let's add a return here (just a small optimization so we won't get into the `switch` below)

urish:
Thanks!

PR #45, opened 2021-05-28
feat(logging): Added basic logging functionality

proteusnet:


src/rp2040.ts
```diff
@@ -1688,27 +1687,21 @@ export class RP2040 {
     else if (opcode === 0b1011111100100000) {
       // do nothing for now. Wait for event!
       this.cycles++;
-      if (this.logMissingInstructions) {
-        console.log('WFE');
-      }
+      this.logger.warn("WFE");
```
urish, line 1690: Those instructions (WFE/WFI/YIELD) should probably be info and not warn, as the code just runs fine without them (at least in the single core variation)

src/utils/logging.ts
```diff
@@ -0,0 +1,57 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+export interface LogInterface {
+    debug(msg: string, ...supportingData: any[]): void;
+    warn(msg: string, ...supportingData: any[]): void;
+    error(msg: string, ...supportingData: any[]): void;
+    info(msg: string, ...supportingData: any[]): void;
+}
+
+export enum LogLevel {
+    debug,
```
urish, line 10: Enum values in TypeScript are [typically written PascalCase](https://www.typescriptlang.org/docs/handbook/enums.html), e.g. `Debug`, `Info`. Let's stick with that

src/utils/logging.ts
```diff
@@ -0,0 +1,57 @@
...
+    info,
+    warn,
+    error
+}
+
+
+export class Log implements LogInterface {
+    public currentLogLevel = LogLevel.info;
+    private exitonerror = true;
```
urish, line 19: Should be `throwOnError` (as it throws on error)

urish: Also, no reason to initialize / declare it here - because you already declare it in the constructor

src/utils/logging.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    constructor(loglevel: LogLevel, exitonerror: boolean) {
```
urish, line 21: I suggest adding `name` for the logger, this will also us to print the peripheral name before each log message, and in the future also configure different log names according to the peripheral name.

also, in the constructor, you can use `private readonly logLevel: LogLevel` to declare an initialize the property (instead of declaring it once in the class and once in the constructor and having to initialize it here).

src/utils/logging.ts
```diff
@@ -0,0 +1,57 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+export interface LogInterface {
+    debug(msg: string, ...supportingData: any[]): void;
+    warn(msg: string, ...supportingData: any[]): void;
+    error(msg: string, ...supportingData: any[]): void;
+    info(msg: string, ...supportingData: any[]): void;
+}
```
urish, line 7: Do we really need the supportingData? I prefer to avoid any types, unless absolutely needed.

src/utils/logging.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export enum LogLevel {
+    debug,
+    info,
+    warn,
+    error
+}
+
+
+export class Log implements LogInterface {
```
urish, line 17: Let's call this class `ConsoleLogger` (because it logs to the console). Then we can call the interface `Logging`.

src/rp2040.ts
```diff
@@ -527,12 +527,13 @@ export class RP2040 {
       this.sio.writeUint32(address - SIO_START_ADDRESS, value);
     } else if (address >= USBCTRL_BASE && address < USBCTRL_BASE + 0x100000) {
       // Ignore these USB writes for now
+      this.logger.info("USB write ignored for now");
```
urish, line 530: Good idea!

urish (changes requested):
Thank you so much for your pull request! This is a very good idea indeed :)

I left a few comments. Also, please format the code with prettier (you can run `npx prettier --write src/**/*.ts`). This should run automatically when you commit, I'm not sure why it didn't (we'll probably look into it in one of the upcoming live streams)

src/rp2040.ts
proteusnet: Done

src/utils/logging.ts
proteusnet: Done

src/utils/logging.ts
proteusnet: Done

src/utils/logging.ts
proteusnet: Done

src/utils/logging.ts
proteusnet: Done, removed

src/utils/logging.ts
proteusnet: Done

src/utils/logging.ts
```diff
@@ -0,0 +1,53 @@
...
+  Warn,
+  Error,
+}
+
+export class ConsoleLogger implements Logging {
+  public currentLogLevel;
+  private throwOnError;
+  private loggerName;
+
+  constructor(name: string, currentLogLevel: LogLevel, throwOnError: boolean) {
```
urish, line 20: You can simplify the code by changing the constructor signature to:

```ts
  constructor(readonly name: string, public currentLogLevel: LogLevel, private throwOnError = false) {
  }
```

And then you can also remove the property declaration in the class. Defining them with `readonly` / `private` / `public` in the constructor is equivalent to defining them as properties in the class and assigning them with the value from the constructor parameter. This is a [nice TypeScript shortcut](https://kendaleiv.com/typescript-constructor-assignment-public-and-private-keywords/).

urish (changes requested):
Thanks for making all those fixes :)

Two more things:
1. I left a comment with suggestion how to simplify the code in the constructor of ConsoleLogger

2. I probably didn't explain myself well last time - the idea was to have a different logger name for each peripheral, so you can easily see for each peripherals where the logs are coming from. I spent some time thinking on it, and one way to do it would be add a name parameter for each of the logger functions, e.g. `info(name: string, message: string): void;`. Then, we can add utility methods to the `BasePeripheral class`, e.g. 
   ```typescript
    info(message: string) { 
      this.rp2040.logger.info(this.name, message);
    }
    ```
    and in the RP2040 class, we can add some constant, let's say `const LOG_NAME='rp2040`;`, and use it like: `this.logger.info(LOG_NAME, 'some message');`.
    This way users will be able to implement their own loggers with the ability to display different logs levels for each peripherals, etc. 
    thoughts? any questions?
3. Kind reminder to format the code with "prettier". If you need any assistance with that, please let me know.

Thank you for spending your time working on this!

src/utils/logging.ts
proteusnet: Fixed, first time doing typescript so happy to learn!

proteusnet:
Also added utility methods and moved the name parameter to the invididual calls, ready for re-review. Triple checked I ran prettier this time too, let me know if its not what you expect.

src/utils/logging.ts
```diff
@@ -0,0 +1,48 @@
...
+    return loglevel >= this.currentLogLevel ? true : false;
+  }
+
+  private formatMessage(name: string, msg: string) {
+    const currenttime = new Date().toLocaleString();
+    return `${currenttime} [${name}] ${msg}`;
+  }
+
+  debug(name: string, msg: string): void {
+    if (this.aboveLogLevel(LogLevel.Debug)) console.debug(this.formatMessage(name, msg));
```
urish, line 31: I prefer to break it into multi-line if statement with curly braces:
```typescript
    if (this.aboveLogLevel(LogLevel.Debug)) {
      console.debug(this.formatMessage(name, msg));
    }
```

The reason is: if you later add something to this line and it becomes longer, prettier will break it into two lines. But it doesn't know to add curly braces, so you end up with something like:

```typescript
    if (this.aboveLogLevel(LogLevel.Debug))
      console.debug(this.formatMessage(name, msg) + 'some extra stuff so the line is longer');
```

Which has a [dangerous bug potential](https://embeddedgurus.com/barr-code/2014/03/apples-gotofail-ssl-security-bug-was-easily-preventable/).

urish: Same for warn() and info()

src/rp2040.ts
```diff
@@ -1688,27 +1690,21 @@ export class RP2040 {
...
+      this.logger.info(LOG_NAME, "WFE");
     }
     // WFI
     else if (opcode === 0b1011111100110000) {
       // do nothing for now. Wait for event!
       this.cycles++;
-      if (this.logMissingInstructions) {
-        console.log('WFI');
-      }
+      this.logger.info(LOG_NAME, "WFI");
```
urish, line 1699: Hmm, this doesn't seem like it was properly formatted with prettier, as our prettier configuration requires single quote single ('string' and not "string")

src/peripherals/peripheral.ts
```diff
@@ -31,15 +31,15 @@ export class BasePeripheral implements Peripheral {
   constructor(protected rp2040: RP2040, readonly name: string) {}
 
   readUint32(offset: number) {
-    console.warn(`Unimplemented peripheral ${this.name} read from ${offset.toString(16)}`);
+    this.warn(`Unimplemented peripheral ${this.name} read from ${offset.toString(16)}`);
```
urish, line 34: optional, but now we can remove the peripheral name from the warning message (the logger will display it anyway)

same for the warning in writeUint32()

src/utils/logging.ts
```diff
@@ -0,0 +1,48 @@
...
+    this.throwOnError = throwOnError;
+  }
+
+  private aboveLogLevel(loglevel: LogLevel): boolean {
+    return loglevel >= this.currentLogLevel ? true : false;
+  }
+
+  private formatMessage(name: string, msg: string) {
+    const currenttime = new Date().toLocaleString();
+    return `${currenttime} [${name}] ${msg}`;
```
urish, line 27: The time is a nice addition, but I think it'll be better without a date, and slightly more useful with milliseconds. So now it is:

```
29/05/2021, 17:30:52 [RP2040] Write to undefined address: 18000008
```

and I think something similar this will be more useful:

```
17:30:52.322 [RP2040] Write to undefined address: 18000008
```

urish (changes requested):
Thanks for following through so far, I think we're almost there!

I've run the changes locally to see how it works, and left a few more comments. 

Also, can you please change gdbserver to use this logging? right now it has its own `DEBUG` variable which controls debug prints, and it also prints an annoying "write .. to .." message which clutters the console. So I think it can be useful to use the logger there as well....

urish:
Meanwhile, I also found out why prettier wasn't running automatically for every commit: 7f47c56998c7f7af4457ae4461d5e143514a9a35.

So for now, you'll still need to run prettier (and eslint) manually for the files in this pull-request, unless you rebase your branch on top of master and `npm install`. If want to rebase and need any help with that, I'm here to assist.

src/utils/logging.ts
proteusnet: Done

src/rp2040.ts
proteusnet: Will fix manually, prettier has been run.

proteusnet: Done

src/peripherals/peripheral.ts
proteusnet: Done

src/utils/logging.ts
proteusnet: Done

proteusnet:
Done, ready for review, i dont have much more time to spend on this, so hopefully its useful as it is.

urish:
Very useful, thank you for working on it!

I merged and added a bunch of small fixes: f7664eced30f99bf87a77bcd9038cdaadbf69fc7, and I'll mention this in the next live stream

Issue #49, opened 2021-05-29
ADDS wrong overflow flag


Turro75:
some test case:
Register	StartValue	Emulator	Silicone
r0  		0x5e56b680	0x5e56b680	0x5e56b680
r6  		0x43960000	0xa1ecb680	0xa1ecb680
pc  		0x100054a8	0x100054aa	0x100054aa
xPSR* 		0x01000000	0x81000000	0x91000000
Flags		 [----] 	 [N---] 	 [N--O]

0x100054a8 <l14_1+4>:	36 18	adds	r6, r6, r0

--------------------------------------------------

Register	StartValue	Emulator	Silicone
r1  		0xbf8d1424	0x7f1a2848	0x7f1a2848
pc  		0x100050de	0x100050e0	0x100050e0
xPSR* 		0xa1000000	0x21000000	0x31000000
Flags		 [N-C-] 	 [--C-] 	 [--CO]

0x100050de <__aeabi_dfcmple_guts+80>:	49 18	adds	r1, r1, r1

--------------------------------------------------

Register	StartValue	Emulator	Silicone
r4  		0xb8308000	0x70610000	0x70610000
pc  		0x10005614	0x10005616	0x10005616
xPSR* 		0x01000000	0x21000000	0x31000000
Flags		 [----] 	 [--C-] 	 [--CO]

0x10005614 <l20_1+8>:	24 19	adds	r4, r4, r4

--------------------------------------------------

Register	StartValue	Emulator	Silicone
r2  		0x74bc8000	0x74bc8000	0x74bc8000
r4  		0x43740000	0xb8308000	0xb8308000
pc  		0x10005610	0x10005612	0x10005612
xPSR* 		0x01000000	0x81000000	0x91000000
Flags		 [----] 	 [N---] 	 [N--O]

0x10005610 <l20_1+4>:	a4 18	adds	r4, r4, r2

--------------------------------------------------

Register	StartValue	Emulator	Silicone
r0  		0x5e56b680	0x5e56b680	0x5e56b680
r6  		0x43960000	0xa1ecb680	0xa1ecb680
pc  		0x100054a8	0x100054aa	0x100054aa
xPSR* 		0x01000000	0x81000000	0x91000000
Flags		 [----] 	 [N---] 	 [N--O]

0x100054a8 <l14_1+4>:	36 18	adds	r6, r6, r0

PR #52, opened 2021-05-30
fix(rp2040): Change ADDSreg to be more coherent to datasheet

Turro75:
close ADDS wrong overflow flag #49

### DIFF
--- a/src/rp2040.ts
+++ b/src/rp2040.ts
@@ -942,14 +942,14 @@ export class RP2040 {
       const Rd = opcode & 0x7;
       const leftValue = this.registers[Rn];
       const rightValue = this.registers[Rm];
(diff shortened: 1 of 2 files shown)


src/rp2040.ts
```diff
@@ -942,12 +942,14 @@ export class RP2040 {
       const Rd = opcode & 0x7;
       const leftValue = this.registers[Rn];
       const rightValue = this.registers[Rm];
+      const unsigned_sum = (leftValue + rightValue) >>> 0;
+      const signed_sum = (leftValue | 0) + (rightValue | 0);
```
urish, line 946: Thanks! Can we please call them `signedSum` and `unsignedSum`?

urish (changes requested):

urish:
Thank you valerio!

PR #53, opened 2021-05-30
fix(rp2040): change STRreg to handle negative offset

### LINKED ISSUE #43: STR instruction fails when using a negative offset
The STR instruction has a form where you can specify base register + offset register, e.g.

```
str r0, [r1, r2]
```

It's possible to use a negative offset value (e.g. -4), and the target address will still be correct as the result should wrap at 32 bits. This doesn't work correctly in the current implementation. The following test case passes on the physical hardware but fails in the emulator:

```ts
  it('should execute a `str r2, [r3, r1]` instruction where r1 + r3 > 32 bits', async () => {
    await cpu.setPC(0x20000000);
    await cpu.writeUint16(0x20000000, opcodeSTRreg(r2, r1, r3));
    await cpu.setRegisters({ r1: -4, r3: 0x20041e50, r2: 0x4201337 });
    await cpu.singleStep();
    expect(await cpu.readUint32(0x20041e4c)).toEqual(0x4201337);
    expect(registers.pc).toEqual(0x20000002);
  });
```

I suspect that a similar bug exists for other instructions that accept a memory address in the form of base register + offset register, e.g. STRH, STRB, LDR, LDRB, LDRH, LDRS, LDRSB.

Turro75:
close #43

### DIFF
--- a/src/rp2040.ts
+++ b/src/rp2040.ts
@@ -1539,7 +1539,7 @@ export class RP2040 {
       const Rm = (opcode >> 6) & 0x7;
       const Rn = (opcode >> 3) & 0x7;
       const Rt = opcode & 0x7;
-      const address = this.registers[Rm] + this.registers[Rn];
+      const address = (this.registers[Rm] | 0) + (this.registers[Rn] | 0);
       if (this.slowIO(address)) {
         this.cycles++;
       }
--- a/src/instructions.spec.ts
+++ b/src/instructions.spec.ts
@@ -1114,6 +1114,16 @@ describe('Cortex-M0+ Instruction Set', () => {
     expect(registers.pc).toEqual(0x20000002);
   });
 
+  it('should execute a `str r2, [r3, r1]` instruction where r1 + r3 > 32 bits', async () => {
+    await cpu.setPC(0x20000000);
+    await cpu.writeUint16(0x20000000, opcodeSTRreg(r2, r1, r3));
+    await cpu.setRegisters({ r1: -4, r3: 0x20041e50, r2: 0x4201337 });
+    await cpu.singleStep();
+    const registers = await cpu.readRegisters();
+    expect(await cpu.readUint32(0x20041e4c)).toEqual(0x4201337);
+    expect(registers.pc).toEqual(0x20000002);
+  });
+
   it('should execute an `str r3, [sp, #12]` instruction', async () => {
     await cpu.setPC(0x20000000);
     await cpu.writeUint16(0x20000000, opcodeSTRsp(r3, 12));


src/rp2040.ts
```diff
@@ -1539,7 +1539,7 @@ export class RP2040 {
       const Rm = (opcode >> 6) & 0x7;
       const Rn = (opcode >> 3) & 0x7;
       const Rt = opcode & 0x7;
-      const address = this.registers[Rm] + this.registers[Rn];
+      const address = (this.registers[Rm] | 0) + (this.registers[Rn] | 0);
```
urish, line 1542: I'm thinking it might be better just to fix this in `readUint32()` / `writeUint32()` / `slowIO()` - always convert the address to unsigned 32-bit (with `>>> 0`) at the beginning of those functions.

This will it will fix all the other places (LDR/STR variations), and will also make sure we won't get negative numbers. I think with the current fix, writing to `0xd0000000` (SIO) will not work correctly, because it'll be converted into a negative number?

urish (changes requested):

src/rp2040.ts
Turro75: changing to this code:

const address = this.registers[Rm] + (this.registers[Rn] | 0);

passes even this test (writing a value to DIV_UDIVIDEND)

it('should execute a `str r2, [r3, r1]` instruction where r1 + r3 > 32 bits', async () => {
    await cpu.setPC(0x20000000);
    await cpu.writeUint16(0x20000000, opcodeSTRreg(r2, r1, r3));
    await cpu.setRegisters({ r1: -4, r3: 0xd0000064, r2: 0x4201337 });
    await cpu.singleStep();
    expect(await cpu.readUint32(0xd0000060)).toEqual(0x4201337);
    expect(registers.pc).toEqual(0x20000002);
  });

Turro75: but this solves
const address = (this.registers[Rm] + this.registers[Rn]) >>> 0;

and even this test passes:

it('should execute a `str r2, [r3, r1]` instruction where r1 + r3 > 32 bits', async () => {
    await cpu.setPC(0x20000000);
    await cpu.writeUint16(0x20000000, opcodeSTRreg(r2, r1, r3));
    await cpu.setRegisters({ r1: 0x80000000, r3: 0x50000060, r2: 0x4201337 });
    await cpu.singleStep();
    expect(await cpu.readUint32(0xd0000060)).toEqual(0x4201337);
    expect(registers.pc).toEqual(0x20000002);
  });

Turro75: so in the end You are always right

PR #54, opened 2021-05-30
fix(rp2040): CMP T1 & T2  flags

Turro75:
also fixes a bug in opcodeCMPregT2 in assembler.ts

fix various missing await in instructions test

prettier refactor of some parts on rp2040.ts


urish:
Amazing, thanks!

PR #57, opened 2021-05-31
fix(rp2040): SUBS alignment to CMP implementation

### LINKED ISSUE #56: again SUBS wrong flags
according to what did in CMP, I tested this case and it fails:

it('should execute a `subs r5, r3, r2` instruction', async () => {
    await cpu.setPC(0x20000000);
    await cpu.writeUint16(0x20000000, opcodeSUBSreg(r5, r3, r2));
    await cpu.setRegisters({ r3: 0 });
    await cpu.setRegisters({ r2: 0x80000000 });
    await cpu.singleStep();
    expect(registers.r5).toEqual(0x80000000);
    expect(registers.N).toEqual(true);
    expect(registers.Z).toEqual(false);
    expect(registers.C).toEqual(false);
    expect(registers.V).toEqual(true);
  });

I'm sending a new PR to (hopefully) definitely close this bug

Turro75:
close #56

### DIFF
--- a/src/instructions.spec.ts
+++ b/src/instructions.spec.ts
@@ -1325,6 +1325,34 @@ describe('Cortex-M0+ Instruction Set', () => {
     expect(registers.V).toEqual(false);
   });
 
+  it('should execute a `subs r5, r3, r2` instruction and set N V flags', async () => {
+    await cpu.setPC(0x20000000);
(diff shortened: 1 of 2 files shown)


urish:
Thank you!

PR #59, opened 2021-05-31
fix(rp2040): change RSBS and SBCS according to CMP/SUBS implementation

### LINKED ISSUE #58: RSBS - SBCS wrong flags
SBCS instruction fails with this test:

  it('should execute a `sbcs r0, r3` instruction', async () => {
    await cpu.setPC(0x20000000);
    await cpu.writeUint16(0x20000000, opcodeSBCS(r0, r3));
    await cpu.setRegisters({ r0: 0, r3: 0x80000000, C: true });
    await cpu.singleStep();
    expect(registers.r0).toEqual(0x80000000);
    expect(registers.N).toEqual(true);
    expect(registers.Z).toEqual(false);
    expect(registers.C).toEqual(false);
    expect(registers.V).toEqual(true);
  });

  it('should execute a `sbcs r0, r3` instruction', async () => {
    await cpu.setPC(0x20000000);
    await cpu.writeUint16(0x20000000, opcodeSBCS(r0, r3));
    await cpu.setRegisters({ r0: 0x80000000, r3: 0x0, C: false });
    await cpu.singleStep();
    expect(registers.r0).toEqual(0x7fffffff);
    expect(registers.N).toEqual(false); 
    expect(registers.Z).toEqual(false);
    expect(registers.C).toEqual(true);
    expect(registers.V).toEqual(true);
  });

  and RSBS 

it('should execute a `rsbs r0, r3` instruction', async () => {
    // This instruction is also calledasync  `negs`
    await cpu.setPC(0x20000000);
    await cpu.writeUint16(0x20000000, opcodeRSBS(r0, r3));
    await cpu.setRegisters({ r3: 0x80000000 });
    await cpu.singleStep();
    expect(registers.r0).toEqual(0x80000000);
    expect(registers.N).toEqual(true);
    expect(registers.Z).toEqual(false);
    expect(registers.C).toEqual(false);
    expect(registers.V).toEqual(true);
  });

  it('should execute a `rsbs r0, r3` instruction', async () => {
    // This instruction is also calledasync  `negs`
    await cpu.setPC(0x20000000);
    await cpu.writeUint16(0x20000000, opcodeRSBS(r0, r3));
    await cpu.setRegisters({ r3: 0x7fffffff });
    await cpu.singleStep();
    expect(registers.r0).toEqual(0x80000001);
    expect(registers.N).toEqual(true);
    expect(registers.Z).toEqual(false);
    expect(registers.C).toEqual(false);
    expect(registers.V).toEqual(false);
  });

the fix is easy, both must be rewritten as CMP/SUBS, a PR with is coming

Turro75:
SBCS reworked to reflect latest SUBS implementation
RSBS (NEGS) is actually a special case of SUBS where leftValue is 0

close #58

### DIFF
--- a/src/instructions.spec.ts
+++ b/src/instructions.spec.ts
@@ -1141,6 +1141,34 @@ describe('Cortex-M0+ Instruction Set', () => {
     expect(registers.V).toEqual(false);
   });
 
+  it('should execute a `rsbs r0, r3` instruction', async () => {
+    // This instruction is also calledasync  `negs`
+    await cpu.setPC(0x20000000);
+    await cpu.writeUint16(0x20000000, opcodeRSBS(r0, r3));
+    await cpu.setRegisters({ r3: 0x80000000 });
+    await cpu.singleStep();
+    const registers = await cpu.readRegisters();
+    expect(registers.r0).toEqual(0x80000000);
+    expect(registers.N).toEqual(true);
+    expect(registers.Z).toEqual(false);
+    expect(registers.C).toEqual(false);
+    expect(registers.V).toEqual(true);
+  });
+
+  it('should execute a `rsbs r0, r3` instruction', async () => {
+    // This instruction is also calledasync  `negs`
+    await cpu.setPC(0x20000000);
+    await cpu.writeUint16(0x20000000, opcodeRSBS(r0, r3));
+    await cpu.setRegisters({ r3: 0x7fffffff });
+    await cpu.singleStep();
+    const registers = await cpu.readRegisters();
+    expect(registers.r0).toEqual(0x80000001);
+    expect(registers.N).toEqual(true);
+    expect(registers.Z).toEqual(false);
+    expect(registers.C).toEqual(false);
+    expect(registers.V).toEqual(false);
+  });
+
   it('should execute a `sbcs r0, r3` instruction', async () => {
     await cpu.setPC(0x20000000);
     await cpu.writeUint16(0x20000000, opcodeSBCS(r0, r3));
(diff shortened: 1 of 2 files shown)


src/instructions.spec.ts
```diff
@@ -1141,6 +1141,34 @@ describe('Cortex-M0+ Instruction Set', () => {
     expect(registers.V).toEqual(false);
   });
 
+  it('should execute a `rsbs r0, r3` instruction', async () => {
+    // This instruction is also calledasync  `negs`
```
urish, line 1145: please remove the word "async" from the comment

src/instructions.spec.ts
```diff
@@ -1141,6 +1141,34 @@ describe('Cortex-M0+ Instruction Set', () => {
...
+    const registers = await cpu.readRegisters();
+    expect(registers.r0).toEqual(0x80000000);
+    expect(registers.N).toEqual(true);
+    expect(registers.Z).toEqual(false);
+    expect(registers.C).toEqual(false);
+    expect(registers.V).toEqual(true);
+  });
+
+  it('should execute a `rsbs r0, r3` instruction', async () => {
+    // This instruction is also calledasync  `negs`
```
urish, line 1159: same here :-)

urish (changes requested):
Thanks! Just a small fix in the comment :)

src/instructions.spec.ts
Turro75: OK no problem, finally I realized that all sub related instructions , even immediate can use the same code, if You agree next commit will include them.
I don't think it is worth adding more test as I expect a duplicated behavior

urish: Yes, that sounds fine!

src/rp2040.ts
```diff
@@ -1640,12 +1649,14 @@ export class RP2040 {
       const Rd = opcode & 0x7;
       const leftValue = this.registers[Rn];
       const rightValue = this.registers[Rm];
-      const result = (leftValue - rightValue) | 0;
+      const result = leftValue - rightValue;
       this.registers[Rd] = result;
-      this.N = (leftValue | 0) < (rightValue | 0);
-      this.Z = leftValue === rightValue;
+      this.N = !!(result & 0x80000000);
+      this.Z = (result & 0xffffffff) === 0;
```
urish, line 1655: Just for your information - `result | 0` would also do the same (shrink the value to 32 bits). But it's also fine this way!

urish:
This got me thinking - since all these functions share the same logic, perhaps it'll be better if way make a function (e.g. subtractUpdateFlags or something like that) that:

1. Takes `leftValue` and `rightValue` as arguments
2. Calculates the result
3. Updates the flag
4. Returns the result

Then, instructions like SUBS will use the result (e.g. store it in a register), while CMP will call this function just to update the flags, without using the result.

What do you think?

Turro75:
that's good, I'll do it Tomorrow before the stream

urish:
Excellent, thanks!

Turro75:
I moved also addition instructions to a dedicated function, created a new PR

PR #60, opened 2021-06-01
fix(rp2040): Moved subs and adds code to a dedicated function

Turro75:
ADCS, ADDS T1/T2, ADDSreg, CMN all use additionUpdateFlags

CMP, CMPreg T1/T2, NEGS/RSBS, SBCS, SUBS T1/T2, SUBSreg all use substractUpdateFlags

### DIFF
--- a/src/rp2040.ts
+++ b/src/rp2040.ts
@@ -880,7 +880,7 @@ export class RP2040 {
     }
   }
 
-  substractUpdateFlags(minuend: number, subtrahend: number) {
+  private substractUpdateFlags(minuend: number, subtrahend: number) {
     const result = minuend - subtrahend;
     this.N = !!(result & 0x80000000);
     this.Z = (result & 0xffffffff) === 0;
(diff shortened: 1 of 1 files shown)


src/rp2040.ts
```diff
@@ -880,6 +880,28 @@ export class RP2040 {
     }
   }
 
+  substractUpdateFlags(minuend: number, subtrahend: number) {
```
urish, line 883: Funny, I didn't know minuend/subtrahend was even a word! I learned something new :)

Can we make this method private?

src/rp2040.ts
```diff
@@ -880,6 +880,28 @@ export class RP2040 {
...
+    this.N = !!(result & 0x80000000);
+    this.Z = (result & 0xffffffff) === 0;
+    this.C = minuend >= subtrahend;
+    this.V =
+      (!!(result & 0x80000000) && !(minuend & 0x80000000) && !!(subtrahend & 0x80000000)) ||
+      (!(result & 0x80000000) && !!(minuend & 0x80000000) && !(subtrahend & 0x80000000));
+    return result;
+  }
+
+  additionUpdateFlags(addend1: number, addend2: number) {
```
urish, line 894: Let's call it `addUpdateFlags()` (because the other one is subtract)

urish (changes requested):
I love how it makes the code shorter! Let's just make the new methods private? (+ rename additionUpdateFlags to addUpdateFlags)

src/rp2040.ts
Turro75: > I didn't know minuend/subtrahend was even a word!
the google's power :-)

urish:
Great, thanks a lot valerio! 

Just published it as  rp2040js 0.6.3

PR #65, opened 2021-06-04
test(PIO):improving `SET PINS` test to be unaffected by other pins

Turro75:
the set pins requires the result and-ed to avoid the other pins status

### DIFF
--- a/src/peripherals/pio.spec.ts
+++ b/src/peripherals/pio.spec.ts
@@ -46,6 +46,8 @@ const INSTR_MEM1 = 0x5020004c;
 const INSTR_MEM2 = 0x50200050;
 const INSTR_MEM3 = 0x50200054;
 const SM0_SHIFTCTRL = 0x502000d0;
+const IN_SHIFTDIR = 18;
+const OUT_SHIFTDIR = 19;
 const SM0_ADDR = 0x502000d4;
 const SM0_INSTR = 0x502000d8;
 const SM0_PINCTRL = 0x502000dc;
@@ -103,11 +105,10 @@ describe('PIO', () => {
     // Clear FIFOs
     await cpu.writeUint32(SM0_SHIFTCTRL, FJOIN_RX);
     //values at reset
-    await cpu.writeUint32(SM0_SHIFTCTRL, (1 << 18) | (1 << 19));
+    await cpu.writeUint32(SM0_SHIFTCTRL, (1 << IN_SHIFTDIR) | (1 << OUT_SHIFTDIR));
     await cpu.writeUint32(SM0_PINCTRL, 5 << SET_COUNT_SHIFT);
   }
 
-  // TODO figure out why this specific test fails on silicone?
   it('should execute a `SET PINS` instruction correctly', async () => {
     // SET PINS, 13
     // then check the debug register and verify that that output from the pins matches the PINS value
--- a/src/peripherals/pio.spec.ts
+++ b/src/peripherals/pio.spec.ts
@@ -61,6 +61,7 @@ const FJOIN_RX = 1 << 30;
 const DBG_PADOUT = 0x5020003c;
 
 const SET_COUNT_SHIFT = 26;
+const SET_COUNT_BASE = 5;
 const OUT_COUNT_SHIFT = 20;
 
 const VALID_PINS_MASK = 0x3fffffff;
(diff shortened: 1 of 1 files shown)


Turro75:
as alternative the GDB_PADOUT can be reset by writing 0 to PINS with a MOV instruction, I don't know if GDB_PADOUT is affected by GPIO status.

```
 async function resetStateMachines() {

    await cpu.writeUint32(CTRL, 0xf0);
    await cpu.writeUint32(SM0_INSTR, pioJMP(PIO_COND_ALWAYS, 0)); // Jump machine 0 to address 0

    // Clear FIFOs
    await cpu.writeUint32(SM0_SHIFTCTRL, FJOIN_RX);  
    await cpu.writeUint32(SM0_SHIFTCTRL, 0);

    //reset Pins/GDB_PADOUT
    await cpu.writeUint32(SM0_INSTR, pioSET(PIO_DEST_X, 0));
    await cpu.writeUint32(SM0_INSTR, pioMOV(PIO_DEST_PINS, PIO_OP_NONE, PIO_SRC_X)); 
  }
```

src/peripherals/pio.spec.ts
```diff
@@ -102,16 +103,21 @@ describe('PIO', () => {
...
     // then check the debug register and verify that that output from the pins matches the PINS value
+    const shiftAmount = 5;
+    const pinValue = 25;
     await resetStateMachines();
-    await cpu.writeUint32(SM0_PINCTRL, 5 << SET_COUNT_SHIFT);
-    await cpu.writeUint32(SM0_INSTR, pioSET(PIO_DEST_PINS, 13));
-    expect(await cpu.readUint32(DBG_PADOUT)).toBe(13);
+    await cpu.writeUint32(SM0_PINCTRL, (5 << SET_COUNT_SHIFT) | (shiftAmount << SET_COUNT_BASE));
+    await cpu.writeUint32(SM0_INSTR, pioSET(PIO_DEST_PINS, pinValue));
+    expect((await cpu.readUint32(DBG_PADOUT)) & (pinValue << shiftAmount)).toBe(
```
urish, line 118: Actually, we want to check all the pins (and not only the ones we've set to 1), so we need the mask to be 0x1f and not pinValue, e.g.:
```
    expect((await cpu.readUint32(DBG_PADOUT)) & (0x1f << shiftAmount)).toBe(
```

urish (changes requested):

urish:
Thank you Valerio!

> I don't know if DBG_PADOUT is affected by GPIO status, anyway for testing purpose seems not important.

The datasheet says: "Read to sample the pad output values PIO is currently driving to the GPIOs.", so as I understand it it's only affected by PIO and not by any external peripherals.

In any case, both methods are fine, I don't have any strong preference for one over the other.

src/peripherals/pio.spec.ts
Turro75: Right, I pushed another commit

Turro75:
> In any case, both methods are fine, I don't have any strong preference for one over the other.

me neither, both work.
a little bit better the reworked test as it's more generic and allows the testing of any combination of `SET PINS` while the other way uses instructions to test instructions, but who cares it's a test and now it passes.

src/peripherals/pio.spec.ts
```diff
@@ -101,17 +102,27 @@ describe('PIO', () => {
     await cpu.writeUint32(SM0_INSTR, pioJMP(PIO_COND_ALWAYS, 0)); // Jump machine 0 to address 0
     // Clear FIFOs
     await cpu.writeUint32(SM0_SHIFTCTRL, FJOIN_RX);
-    await cpu.writeUint32(SM0_SHIFTCTRL, 0);
+    //values at reset
+    await cpu.writeUint32(SM0_SHIFTCTRL, (1 << 18) | (1 << 19));
+    await cpu.writeUint32(SM0_PINCTRL, 5 << SET_COUNT_SHIFT);
   }
 
   // TODO figure out why this specific test fails on silicone?
```
urish, line 110: We can also remove this TODO comment (it doesn't fail anymore!)

urish (changes requested):
Awesome! Just one small request :)

src/peripherals/pio.spec.ts
```diff
@@ -101,17 +102,27 @@ describe('PIO', () => {
     await cpu.writeUint32(SM0_INSTR, pioJMP(PIO_COND_ALWAYS, 0)); // Jump machine 0 to address 0
     // Clear FIFOs
     await cpu.writeUint32(SM0_SHIFTCTRL, FJOIN_RX);
-    await cpu.writeUint32(SM0_SHIFTCTRL, 0);
+    //values at reset
+    await cpu.writeUint32(SM0_SHIFTCTRL, (1 << 18) | (1 << 19));
```
urish, line 106: What are do bit 18 / 19 mean?

Let's define a constant for them (like we have `FJOIN_RX`), so the test code is easier to understand

urish:
Thanks!

PR #87, opened 2021-11-03
Fix `bitReverse` function in `pio.ts`

martin2250:
pretty self-explanatory, this is another bug I found while playing around with the PIOs.

https://jsfiddle.net/bwn6j75r/

Cheers!

### DIFF
--- a/src/peripherals/pio.ts
+++ b/src/peripherals/pio.ts
@@ -81,11 +81,11 @@ export enum WaitType {
 }
 
 function bitReverse(x: number) {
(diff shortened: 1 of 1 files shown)


urish:
Good catch!

You can blame me for [copying this code from StackOverflow](https://youtu.be/HUW8q__jqr4?list=PLLomdjsHtJTxT-vdJHwa3z62dFXZnzYBm&t=8548) without testing it thoroughly. Never blindly trust stackoverflow!

urish:
Fix is up! thanks

martin2250:
> copying this code from StackOverflow

caught in the act 😉 
thanks!

Issue #88, opened 2021-11-19
MicroPython: Filesystem will not mount


DevyHeavy:
rp2040js: 0.14.6
MicroPython: 1.17
OS: Windows 10

Steps to reproduce:

1. clone rp2040js: `git clone https://github.com/wokwi/rp2040js.git`
2. `cd rp2040js`
3. `npm install`
4. `npm run start:micropython`
5. wait for repl: `f = open('test.txt', 'w')`
6. prints: 
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
OSError: [Errno 19] ENODEV
```
Corresponds to filesystem not mounted: https://forum.micropython.org/viewtopic.php?t=1855#p10444

urish:
The file system is not initialized by the emulator. If you want to have a file system, you'd need to create it yourself and make it available in a specific location in the simulated flash memory. 

For MicroPython on the Raspberry Pi Pico, that'd be [littlefs](https://github.com/littlefs-project/littlefs) at flash offset 0xa0000, with a block size of 4k and 352 blocks.

What project are you working on?

DevyHeavy:
OK,

So I can use littlefs to create a filesystem starting at 0xa0000 for 352 blocks at 4kB per block.

I have been able to successfully write files that MicroPython can see using the code you provided before: https://github.com/wokwi/rp2040js/issues/81#issuecomment-932823793; however, I still get `Errno 19` when creating fiels in MicroPython (step 5).

Here are the steps I use that allows MicroPython see files but still fails on MP file creation:
1. Set each file in `flash` using littlefs.wasm
2. copy `flash` to emulator flash

Here's the code for step 2:
```
console.log("Flash FS copy started");

littlefs._lfs_unmount(lfs);
littlefs._free(lfs);
littlefs._free(config);

rp2040.flash.set(flash, 0xa0000);

console.log("Flash FS copy ended");
```

urish:
I thought your name was familiar :)

For proper writing to flash, you'd need to implement the SSI peripheral (see section 4.10 of the datasheet) 

A workaround would be to patch the `flash_range_program` function from the bootrom. MicroPython uses that function to write to the Flash.

The details of that function are available in section 2.8.2.1.3. 

![image](https://user-images.githubusercontent.com/892318/142689715-9c2f2cdc-0625-49cf-a9c7-5a8a44e79a06.png)

[...]

Issue #81, opened 2021-10-01
Using 'import' in MicroPython in the browser


DevyHeavy:
Hi again,

I am looking into how one might get 'import' to work in MicroPython in the browser, also without fetching it from a server.

For example, say I have the following python files to be used in MicroPython

main.py
```
import test
print(test.value)
```

test.py
```
value = 10
```

the expected output is '10'.

On a filesystem, this project would look like
* ProjectDir
   * main.py
   * test.py

Now, say a user makes both of these files on some kind of text editor on a webpage and there is access to the contents of each file through javascript. How could 'main.py' know about 'test.py' in the browser?

I know of BrowserFS (https://github.com/jvilk/BrowserFS) but am not entirely sure how this would work with rp2040js.

I have seen ways of overriding the MicroPython importer to use provided modules to fetch the contents of files from a URL (https://github.com/micropython/micropython/issues/4972#issuecomment-626111144). I suppose that may be used in some way by using a Blob (https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL) to provide a client-side URL for the file contents to be fetched from in the MicroPython script (?)

In any case, I am looking for insight on how to get 'import' working in python scripts with MicroPythin using this library in browser

DevyHeavy:
Just tested 'file = open("data.txt", "w")' to try and create a file in MicroPytho, it returns Erno 19 which means filesystem not mounted, as expected

DevyHeavy:
I think I'm starting to figure it out, at least where to maybe start. Looks like the simulator uses littlefs.wasm: https://www.npmjs.com/package/littlefs

urish:
Good observation - MicroPython indeed uses LittleFS. 

On the Raspberry Pi Pico, I found out the it works this way:
1. The file system starts at offset 0xa0000 of the flash
2. The block size is 4096
3. The amount of blocks is 352

You can also deduce these values from the MicroPython source code:
1. [Filesystem size](https://github.com/micropython/micropython/blob/d42cba0d22cac812cc5a12f4670010b45932eafa/ports/rp2/boards/PICO/mpconfigboard.h#L3-L3)
2. [Filesystem offset](https://github.com/micropython/micropython/blob/d42cba0d22cac812cc5a12f4670010b45932eafa/ports/rp2/rp2_flash.c#L42-L42)
[...]

PR #90, opened 2022-01-07
feat(ADC): support DMA by setting DREQs based on ADC FIFO fill level

tomods:
The existing ADC and DMA code already had nearly everything – so this just adds some DREQ logic to the ADC code. It is inspired by the PIO DREQ code.

I did some basic tests from Micropython, using the approach from https://iosoft.blog/2021/10/26/pico-adc-dma/ – seems to work fine so far.

Please note that I'm a Javascript/Typescript novice ;)

### DIFF
--- a/src/peripherals/adc.ts
+++ b/src/peripherals/adc.ts
@@ -1,6 +1,7 @@
 import { IRQ } from '../irq';
 import { RP2040 } from '../rp2040';
 import { FIFO } from '../utils/fifo';
+import { DREQChannel } from './dma';
 import { BasePeripheral, Peripheral } from './peripheral';
 
 const CS = 0x00; // ADC Control and Status
@@ -96,6 +97,7 @@ export class RPADC extends BasePeripheral implements Peripheral {
   };
 
   readonly fifo = new FIFO(4);
+  readonly dreq = DREQChannel.DREQ_ADC;
 
   // Registers
   cs = 0;
@@ -156,6 +158,17 @@ export class RPADC extends BasePeripheral implements Peripheral {
     this.onADCRead(this.activeChannel);
   }
 
+  private updateDMA() {
+    if (this.fcs & FCS_DREQ_EN) {
+      const thres = (this.fcs >> FCS_THRESH_SHIFT) & FCS_THRES_MASK;
+      if (this.fifo.itemCount >= thres) {
+        this.rp2040.dma.setDREQ(this.dreq);
+      } else {
+        this.rp2040.dma.clearDREQ(this.dreq);
+      }
+    }
+  }
+
   completeADCRead(value: number, error: boolean) {
     this.busy = false;
     this.result = value;
@@ -178,6 +191,7 @@ export class RPADC extends BasePeripheral implements Peripheral {
           value |= FIFO_ERR;
         }
         this.fifo.push(value);
+        this.updateDMA();
         this.checkInterrupts();
       }
     }
@@ -228,7 +242,9 @@ export class RPADC extends BasePeripheral implements Peripheral {
           this.fcs |= FCS_UNDER;
           return 0;
         } else {
-          return this.fifo.pull();
+          const value = this.fifo.pull();
+          this.updateDMA();
+          return value;
         }
       case DIV:
         return this.clockDiv;


urish:
Looks good, thanks! 

Mind sharing here the complete test code that you used?

tomods:
One thing I noticed is that the sample timing using `mcu.clock.createTimer()` seems to be very inaccurate – like _much _ longer than the configured 2 µs. 

I tested this using `process.hrtime.bigint()` and got times of around 9 milliseconds. Do you have any idea how this could be improved? I might look into this over the weekend – unless I receive my Pico HW tomorrow ;)

tomods:
> Mind sharing here the complete test code that you used?

Unfortunately, it's currently not in a good shape at all – like a combination of Python and Typescript hacks. But I plan to publish my (Python) project as open source :)

tomods:
Another thing I noticed – I think the simulated ADC code is in need of some better reset logic. Currently, it cannot get out of `busy` without actually completing an ADC read. I might also look into this soon.

urish:
> Unfortunately, it's currently not in a good shape at all – like a combination of Python and Typescript hacks. But I plan to publish my (Python) project as open source :)

The reason I'm asking - if at some point in the future there someone points out issues with ADC + DMA, we'll have a baseline code that we know works well.

> One thing I noticed is that the sample timing using mcu.clock.createTimer() seems to be very inaccurate – like much longer than the configured 2 µs.

Yes, if you need timing accuracy, you should replace the clock with a virtual one. The real time clock can never be accurate enough on a preemptive multi-tasking OS (it might be possible if you pin the process to a specific core, and make sure nothing else runs on the same core, but I doubt anyone will do this).

urish:
Also, I'm interested: how have you come across rp2040js? Do you have some ideas to build upon it?

tomods:
> The reason I'm asking - if at some point in the future there someone points out issues with ADC + DMA, we'll have a baseline code that we know works well.
OK, makes sense, I'll somehow compile what I did. Not sure what would be the best way to document this – just make it part of this PR? But I probably won't put up much effort for automatizing such a test.

> Yes, if you need timing accuracy, you should replace the clock with a virtual one.
That sounds like a good idea. But I'm not sure how I would go about that – and not sure I'm currently interested in digging into this :)

> The real time clock can never be accurate enough on a preemptive multi-tasking OS (it might be possible if you pin the process to a specific core, and make sure nothing else runs on the same core, but I doubt anyone will do this).
Well, I'm wondering if there are some "low hanging fruit" for this problem – something like e.g. https://github.com/onury/tasktimer
I mean, it doesn't have to be extremely accurate, just somewhat closer.
In general, with a modern Linux kernel, it should be possible to get much better accuracy. However, I'm not sure how well Node lends itself to something like that.

> Also, I'm interested: how have you come across rp2040js?
Um, not completely sure any more – it either was by just googling some Pico DMA/ADC details, or by actually searching for something like "raspberry pi pico simulator" because my Pico HW is not arriving and I wanted to test some code ;) What sold me on it was the example https://wokwi.com/arduino/projects/300504213470839309 and of course the Github repo. 
BTW – I think those Micropython examples are not reachable directly via the wokwi.com page, only from the `README.md`.

Definitely a very pleasant surprise that you took the effort of creating this!

> Do you have some ideas to build upon it?
Well, not too much currently to be honest. I'm just using it to run my code in an actually realistic Micropython environment, so for my developing and testing my project code. I find it very impressive that you can run the "official" Micropython port/SDK.
In my test description for ADC/DMA I will also include how I connected the "usual" Micropython development tools to rp2040js. Could be worth it to try and include that in your `demo` scripts – I'm fairly sure that this could be done directly from Typescript instead of attaching `socat` to stdin/stdout ;)

tomods:
OK, I now don't think it would be easy at all to use a better "real" clock here...

urish:
>  just make it part of this PR? But I probably won't put up much effort for automatizing such a test.

Thanks! Adding a comment with the code in this discussion can be very useful for anyone who ever touches this code in the future. It's okay not automating the test - we don't have automated tests for all the peripherals.

> (BTW – I think those nice Micropython examples are not reachable directly via the wokwi.com page, only from the README.md.)

Good point - I opened wokwi/wokwi-features#265 to track this. And thanks for the feedback!

> OK, I now don't think it would be easy at all to use a better "real" clock here...

If you are interested, I can probably come up with an example. It's would probably be similar in nature to the [clock implementation we use in AVR8js](https://github.com/wokwi/avr8js/blob/8c1f76fb3680c05030abfbf5e8769698f934cbac/src/cpu/cpu.ts#L176-L230)

tomods:
Alright, regarding ADC/DMA testing –

Everything I did is based on my project which I just published on https://github.com/tomods/GrinderController.
The code links I'm adding in this comment are permalinks including the current commit id, so they hopefully won't get broken in the future.

ADC/DMA handling is done in [RP2040ADC.py](https://github.com/tomods/GrinderController/blob/f8ee6a11a23d23d857b0f1b8e0509827355b8d72/RP2040ADC.py). The state machine calls `capture_start()` and `wait_and_read_average_u12()` every cycle.

To provide some input to the simulated ADC, I just hacked up your demo script, see [micropython-run.ts](https://github.com/tomods/GrinderController/blob/f8ee6a11a23d23d857b0f1b8e0509827355b8d72/tools/micropython-run.ts). This is probably not very elegant.

I then used `socat` to start the simulation and attach its stdin/stdout to a "virtual serial port", see [start_socat_rp2040js.sh](https://github.com/tomods/GrinderController/blob/f8ee6a11a23d23d857b0f1b8e0509827355b8d72/tools/start_socat_rp2040js.sh). This way, I can just use Micropython's `pyboard.py` or similar tools to start the scripts.

As rp2040js does not support flash/filesystem (yet), I used one more hack – just removing all Python `import`s internal to my project, and run `pyboard.py -d ~/git/rp2040js/serialport rp_devices.py RP2040ADC.py grinder_controller.py grinder_controller_states.py grinder_debouncer.py grinder_hardware.py main.py` to actually run my code.

The debug logging my code currently does will then output the averaged ADC values coming in via DMA every 10 cycles. So then I just looked at those and manually compared them to the values provided in my `micropython-run.ts` – and they are in fact plausible :)

Before applying the changes, DMA would – quite obviously – hang indefinitely. Ideally, this test would be extended to really test that each and every ADC value is read/copied, but I currently don't see a big risk here.

You can also find my timing test code using `process.hrtime.bigint()` hacked into `micropython-run.ts`. It just dumps the time between consecutive `adc.onADCRead()` calls into a file.

urish:
Alright, this seems to work well. Thanks Tobias!

Released as rp2040js 0.14.8, also made a Wokwi test project based on your code:

https://wokwi.com/arduino/projects/320232834674459219

tomods:
Nice, you're welcome!

By the way – did I see correctly that your simulator web "frontend" is not open source? At least I couldn't find it. Would be interested in learning how you did the file handling – especially for the Python scripts – and how the simulated inputs/output are connected to rp2040js.

urish:
> did I see correctly that your simulator web "frontend" is not open source?

True

> how you did the file handling

There are some details here (for reading):

https://github.com/wokwi/rp2040js/issues/81#issuecomment-932823793

And if you wish to make the virtual flash filesystem writable as well, check out the discussion in #88

urish:
p.s. [GrinderController](https://github.com/tomods/GrinderController) sounds like a cool project. 

Instead of measuring the battery voltage, how about measuring the current? This can be done directly (by adding a low-resistance power resistor in series), or indirectly (with an hall-effect current sensor)

tomods:
Thanks for checking out my project :)

Measuring the current – more or less directly – was my first thought for the project, and is currently my "backup plan". Basically, I'd like to keep things as simple as possible, especially regarding hardware components.

So, my current plan is to try it out using the battery voltage measurement only – if that fails, I will probably try to add a shunt resistor to measure the current.

tomods:
> There are some details here (for read-only access):
> 
> [#81 (comment)](https://github.com/wokwi/rp2040js/issues/81#issuecomment-932823793)

That worked :)

Would you be interested in a PR for `micropython-run.ts` that allows loading a littefs image file into the simulated flash? I decided on creating the image file outside of the Typescript world, and then just read this. But this could later be "upgraded" to actually generate the image file from Typescript.

urish:
Perhaps it's better to have a different example, `micropython-fs.ts` or so which allows loading an image into the simulated flash, as it's a bit more advanced use caes. Other than that, such PR is welcome (you can also mention the issue in a comment, for anyone who want to create the image directly in TypeScript)

tomods:
(littlefs stuff continued at https://github.com/wokwi/rp2040js/pull/92)

PR #91, opened 2022-01-08
feat(DMA): implement DMA sniffing (partially)

tomods:
Implements (parts of) the RP2040 datasheet, chapter 2.5.5.2.

Currently, the code disobeys some settings in SNIFF_CTRL.
It just assumes OUT_INV = OUT_REV = SNIFF_BSWAP = 0; CALC = 0xf.
I.e., it can only do simple summation, no further checksumming.

Is the style OK? Would you be willing to merge this without the features mentioned above? I'm not really in the mood to add the other checksumming options...

Tested with a transfer size of 16 bits using the code you already know from https://github.com/wokwi/rp2040js/pull/90 ;)
Just with the commented-out parts of [RP2040ADC.py](https://github.com/tomods/GrinderController/blob/f8ee6a11a23d23d857b0f1b8e0509827355b8d72/RP2040ADC.py) enabled. So you can easily test this in https://wokwi.com/arduino/projects/320232834674459219

### DIFF
--- a/src/peripherals/dma.ts
+++ b/src/peripherals/dma.ts
@@ -120,6 +120,15 @@ const EN = 1 << 0;
 const CHn_CTRL_TRIG_WRITE_MASK = 0xffffff;
 const CHn_CTRL_TRIG_WC_MASK = READ_ERROR | WRITE_ERROR;
 
+// SNIFF_CTRL bits
+const OUT_INV = 1 << 11;
+const OUT_REV = 1 << 10;
+const SNIFF_BSWAP = 1 << 9;
+const CALC_MASK = 0xf;
+const CALC_SHIFT = 5;
+const DMACH_MASK = 0xf;
+const DMACH_SHIFT = 1;
+
 export class RPDMAChannel {
   private ctrl = 0;
   private readAddr = 0;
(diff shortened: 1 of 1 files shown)


urish:
Thanks!

Would it be hard to support OUT_INV, OUT_REV, and SNIFF_BSWAP as well?

For the bit reversal part, we already have [an implementation](https://github.com/wokwi/rp2040js/blob/85fdd98c7cbb2c66f1ad51d4ea739c7927668c61/src/peripherals/pio.ts#L83). You could move it to `src/utils/bits.ts`, sharing it between the PIO and DMA.

tomods:
Nah, not hard at all. Will look into it when I have some more time

src/peripherals/dma.ts
```diff
@@ -549,4 +587,13 @@ export class RPDMA extends BasePeripheral implements Peripheral {
     this.rp2040.setInterrupt(IRQ.DMA_IRQ0, !!this.intStatus0);
     this.rp2040.setInterrupt(IRQ.DMA_IRQ1, !!this.intStatus1);
   }
+
+  sniffTransfer(channel: RPDMAChannel, datum: number) {
+    const sniffChannel = (this.sniffCtrl >> DMACH_SHIFT) & DMACH_MASK;
+    if ((this.sniffCtrl & EN) && (channel.index === sniffChannel)) {
+      // TODO: Actually obey settings in sniffCtrl...
+      // currently assuming OUT_INV = OUT_REV = SNIFF_BSWAP = 0; CALC = 0xf
+      this.sniffData += datum;
```
urish, line 596: Probably better to cut at 32 bits (in JavaScript, all numbers are stored as double precision floating point by default):

```js
this.sniffData = (this.sniffData + datum) >>> 0;
```

tomods: Oops, good catch. I didn't think about this at the moment – for the ADC/DMA register fiddling from Micropython, `uctypes` did this for me ;)

urish:
Guess you are busy nowadays?

PR #96, opened 2022-05-03
feat(sio): Implement Interpolator

Ferdi265:
This PR implements the "Interpolator" peripheral in the SIO block.

The implementation of the Interpolator is ported from my [rp2040-interpolator](https://github.com/Ferdi265/rp2040-interpolator) implementation, which I tested against a real PICO.

This initial implementation *should* work, though I have not tested it thoroughly yet, since my nice automated test case generation tool for testing this is not working yet.

I have tested the examples from the datasheet, and those work as expected.

As the Interpolator is quite a chunk of code, it might maybe also be a good idea to put it into its own file.

This PR is currently a bit WIP, and I'm open to feedback and suggestions regarding implementation and code style.


src/sio.ts
```diff
@@ -33,11 +33,234 @@ const DIV_QUOTIENT = 0x070; //  Divider result quotient
...
 
+class InterpolatorConfig {
+  shift = 0;
+  mask_lsb = 0;
+  mask_msb = 0;
+  is_signed = false;
+  cross_input = false;
+  cross_result = false;
+  add_raw = false;
+  force_msb = 0;
```
urish, line 83: The naming convention is to use `camelCase`, so please change the member names accordingly (e.g. maskLSB, isSigned, etc).

src/sio.ts
```diff
@@ -33,11 +33,234 @@ const DIV_QUOTIENT = 0x070; //  Divider result quotient
...
+}
+
+class BitUtil {
+    static s32(n: number) {
+        return n | 0;
+    }
+
+    static u32(n: number) {
+        const s32 = BitUtil.s32;
+        return s32(n) < 0 ? s32(n) + 0x100000000 : s32(n);
```
urish, line 135: A simpler, more performant way to implement this:  `n >>> 0`

src/sio.ts
```diff
@@ -33,11 +33,234 @@ const DIV_QUOTIENT = 0x070; //  Divider result quotient
...
+      ((Number(this.clamp)        & 1)        << 22) |
+      ((Number(this.overf0)       & 1)        << 23) |
+      ((Number(this.overf1)       & 1)        << 24) |
+      ((Number(this.overf)        & 1)        << 25) |
+      ((this._reserved0           & 0b111111) << 26)
+    );
+  }
+}
+
+class BitUtil {
```
urish, line 128: Please move these two functions into [src/utils/bit.ts](https://github.com/wokwi/rp2040js/blob/master/src/utils/bit.ts), then import them from that file. Also, since they are totally pure and stateless, no need to wrap them in a class.

src/sio.ts
```diff
@@ -33,11 +33,234 @@ const DIV_QUOTIENT = 0x070; //  Divider result quotient
...
+    this.smresult1 = u32(result1);
+    this.result0 = u32(do_blend ? alpha1 : (do_clamp ? clamp0 : addresult0) | (ctrl0.force_msb << 28));
+    this.result1 = u32((do_blend ? blend1 : addresult1) | (ctrl0.force_msb << 28));
+    this.result2 = u32(addresult2);
+
+    ctrl0.overf0 = overf0;
+    ctrl0.overf1 = overf1;
+    ctrl0.overf = overf;
+    this.ctrl0 = u32(ctrl0.toUint32());
+    this.ctrl1 = u32(ctrl1.toUint32());
```
urish, line 225: It makes sense to move the `u32` call into `toUint32()`, since the function name implies that it returns an unsigned int.

src/sio.ts
```diff
@@ -33,11 +33,234 @@ const DIV_QUOTIENT = 0x070; //  Divider result quotient
...
+        return n | 0;
+    }
+
+    static u32(n: number) {
+        const s32 = BitUtil.s32;
+        return s32(n) < 0 ? s32(n) + 0x100000000 : s32(n);
+    }
+}
+
+class Interpolator {
```
urish, line 139: Let's move this class into a separate file, as you suggested in the description of this pull request.

src/sio.ts
```diff
@@ -33,11 +33,234 @@ const DIV_QUOTIENT = 0x070; //  Divider result quotient
...
+  cross_input = false;
+  cross_result = false;
+  add_raw = false;
+  force_msb = 0;
+  blend = false;
+  clamp = false;
+  overf0 = false;
+  overf1 = false;
+  overf = false;
+  _reserved0 = 0;
```
urish, line 89: Do we really need to store the reserved bits? I think discarding them and just returning them as zeroes makes more sense.

urish (changes requested):
Thank you so much Fredi!

I've left some comments, mostly about the code style. Overall it looks great!

Also, I'm curious, what sparked your interest in implementing the interpolator?

Ferdi265:
Hi! Thanks for the quick answer on the PR!

The Review comments all make perfect sense, will address them once I get to my laptop :)

I mainly looked at the Interpolator because I want to write a demo for the Raspberry Pi Pico, and while reading the datasheet I noticed some very tongue-in-cheek notes about the interpolator being able to do some things "by sheer coincidence". That led me to experimenting with it and the best way to understand it is to implement a software simulation of it.

src/sio.ts
Ferdi265: that's an artifact of this being a direct port of my original interpolator simulator. I initially experimented with setting nonzero values into here and seeing how the interpolator behaves, in the end it just ignores them as one would expect.

Will remove this in the next commit.

src/sio.ts
Ferdi265: sure :)

src/sio.ts
Ferdi265: ahh! yeah, dealing with bit operations in JS definitely was the biggest challenge while porting this. Thanks for the hint!

src/sio.ts
Ferdi265: oh yes, that definitely makes sense

src/sio.ts
Ferdi265: will do!

Ferdi265:
Also, another small question I had while reading `sio.ts`: Is there a separate instance of SIO for each Cortex M0, or do we need to implement the doubled up Hardware Divider / Interpolator manually? (so each CPU sees their own ones as documented in the datasheet)

src/sio.ts
Ferdi265: fixed

src/sio.ts
Ferdi265: fixed

src/sio.ts
Ferdi265: fixed

src/sio.ts
Ferdi265: I noticed the u32 is not necessary at all.
due to the bit operations in InterpCtrl, all resulting values are positive anyway (bit 31 is part of the reserved bits), meaning the u32 call is redundant

src/sio.ts
Ferdi265: fixed

src/sio.ts
Ferdi265: fixed

Ferdi265:
All of your issue comments should be addressed as of 636ac2d.

As for tests: I intend to add an `interpolator.spec.ts` once I have exported a reasonable set of test cases from real hardware (and I'll use these same tests to verify against my other implementations of the interpolator).

src/sio.ts
Ferdi265: done

urish:
> Or is the second core not implemented yet? (I don't quite see it at first glance)

It's not implemented yet (and so far, no one complained!). But once it will be, we'll probably have a different SIO instance per core. We would also need to implement the SIO Inter-processor FIFOs at that point.

urish:
Released as rp2040js 0.16.0, and also updated wokwi.com with the latest code. Here's the interp example from the SDK examples repository:

https://wokwi.com/projects/330747140148036178

I don't have the real hardware with me, so I can't compare the output to the real hardware.

Ferdi265:
Hardware Test results:

- 9 times: OK
- masking: OK
- with sign extend: OK
- result crossover: OK
- blend 1: OK
- blend 2 signed: OK
- blend 2 unsigned: some off-by-ones (need to check and fix that)
- blend 3: OK
- clamp: OK
- linear interp: one off-by-one (probably same bug as blend)
- texture mapping: OK

Looks Like I have a small bug in the precision of the blend mode implementation; I'll look into that ASAP.

Ferdi265:
Looks like the bug is that my rounding is incorrect

urish:
Thanks! Most of the tests are passing on the first attempt, that's very impressive.

Ferdi265:
Well, for the initial implementation in C++, I had to tinker with it for about a day to get it to work, so for the JS port I mostly had everything done, so the bugs here are just from porting to JS's limited bit operations.

PR #102, opened 2022-11-21
Fix wrong value for dreqRx0

mingpepe:


urish:
Good catch, thanks!

Are you using the simulator to develop PIO programs?

mingpepe:
Yes, this emulator is amazing.

PR #103, opened 2022-11-21
SPI busy flag may cause deadlock

mingpepe:
If user call `completeTransmit` in `onTransmit` event, the `busy` flag will be set `false` then `true`. Which may cause emulator waiting for it.

### DIFF
--- a/src/peripherals/pio.ts
+++ b/src/peripherals/pio.ts
@@ -95,7 +95,7 @@ function irqIndex(irq: number, machineIndex: number): number {
 }
 
 const dreqRx0 = [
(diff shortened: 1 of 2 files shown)


urish:
Thanks again! 

I'm curious, what PIO project are you working on?

mingpepe:
To implement camera interface. Not only PIO but trying to run my pico code on this emulator, but seems still have some issues to solve.

urish:
Which issues, for instance?

mingpepe:
I will create new issue later.

PR #104, opened 2022-11-22
USB read delay implementation is missing

mingpepe:


urish:
Does it fix some problem for you?

Because the read delay is currently implemented by `endpointReadDone()`:

https://github.com/wokwi/rp2040js/blob/cac702d96ba358d73da4226b6259488c4671fd59/src/peripherals/usb.ts#L215-L219

mingpepe:
Oh, I miss the detail in `endpointReadDone`. Just think it is missing. This PR should be closed.

One of issues is small packet (size < max_packet_size) not received.
After rp2040 write to DPRAM, the emulator indicate buffer ready immediately, the timing is obviously different from real chip. Though my code should not assume about timing.
https://github.com/wokwi/rp2040js/blob/cac702d96ba358d73da4226b6259488c4671fd59/src/peripherals/usb.ts#L199-L209

`this.rp2040.clock.createTimer` for `this.indicateBufferReady` solved my problem, but finally I decided to modify my code since the timing assumption is not good. But maybe a delay for `this.indicateBufferReady` can be considered and seems like it can combine with `this.writeDelayMicroseconds`.

urish:
Do you mind testing if

mingpepe:
Sadly, this does not fix it. After adding a timer, even the ">>>" prompt not shown.

urish:
Bummer! Did you try to send Ctrl+C or a newline? Sometimes these are required to drop into the prompt

mingpepe:
Yes, it's still not working.

Issue #109, opened 2022-11-28
Core1 support


mingpepe:
Is any plan to do this?

urish:
Seems like Core1 is not widely used - other than you, I think only one person asked about it to date.

Do you have any concrete use case for simulating Core1?

mingpepe:
I want to emulate my code and `core1` is part of it.

Indeed not many projects use `core1` or just use it to do trivial task, but I think `core1` is important if you care about performance.

It's ok not having a schedule for this because I am working in progress but it's not a trivial task and some details I do not understand. I will post what I have done and problems later, some suggestions may be helpful.

urish:
Got it.

We did some preparation at 29746d1e02ff04680809dde23981a351bdd7826c. I think the next step would be:

1. Change [core](https://github.com/wokwi/rp2040js/blob/1d4c96f823ab5f4cef21331a2ef7e103b5736646/src/rp2040.ts#L51) into an array of two cores. Then in `step()`, call executeInstruction on each of the cores. Also in `reset()`, reset both cores.
2. Implement SIO FIFOs (so the two cores can talk), and make sure each core gets its own cpuid, interpolator and integer divider units: 
![image](https://user-images.githubusercontent.com/892318/204468969-65783bc9-d98c-402d-b211-c470b3d54625.png)
3. Making sure that interrupts are routed to the right core. In particular, GPIO has core-specific interrupts (and there's also SIO_IRQ_PROC0/SIO_IRQ_PROC1 which I'm not sure about)

That's from the top of my head.

urish:
Regarding the popularity of this feature: I've opened it for voting back in August. I just checked the [votes status](https://wokwi.com/features?sort=newest#feature-1325417923), seems like nobody voted for it so far.

mingpepe:
:smiling_face_with_tear:

PR #112, opened 2022-12-12
Implement core1

mingpepe:
#109

Test ok with [pico-example code/multicore](https://github.com/raspberrypi/pico-examples/tree/master/multicore)

Interpolator/Divider/Spinlock are not tested or implemented since I have no suitable code to verify.


urish:
Thank you! 

These is a pretty list big of changes, so it'll probably take me some time to go over everything.

Regarding Interpolator/Divider - as far as I remember, the only thing that we need is to have different one per core, so if two cores are using the divider / interpolator at the same time, they don't interfere with eachother:

image

So ideally, we'd want a set of `Interpolator` per core. It might make more sense to also move the divider logic into its own object, and then it'd be easier to create a dedicated copy of it for each core.

urish:
Thanks again! Is this ready for review or not yet?

mingpepe:
Yes, it's ready.

urish:
Now I just need to find a few hours to go through it :)

src/rp2040.ts
```diff
@@ -344,31 +367,58 @@ export class RP2040 {
...
     this.executeTimer = null;
-    this.stopped = false;
-    for (let i = 0; i < 100000 && !this.stopped && !this.core.waiting; i++) {
-      this.core.executeInstruction();
+    for (let i = 0; i < 1000 && !this.core1.stopped && !this.core1.waiting; i++) {
+      this.core1.executeInstruction();
     }
-    if (!this.stopped) {
-      this.executeTimer = setTimeout(() => this.execute(), 0);
+    if (this.core0.stopped || this.core0.waiting) {
```
c1570, line 401: This will make the emulator execute 1000 instructions from core 0, then 1000 instructions from core 1, etc. - the timing between both cores will be totally off, and depending on the actual instructions executed, one core might run much faster than the other core (as instructions may take more than 1 cycle).

It's probably better to keep one execute() function with something like this:
```
  private execute() {
    ...
    for (let i = 0; i < 1000; i++) {
      this.core0.executeInstruction();
      while(this.core1.cycles < this.core0.cycles) {
        this.core1.executeInstruction();
      }
    }
    ...
  }
```

This is still missing the waiting/stopped logic though.

mingpepe: You are right. Had updated.

mingpepe:
@urish  Do you have time to check this PR?

urish:
I added it to my todo list, I will make an effort to allocate some time this week or the next week. 

Ideally, I'd also love to have the multicore test running as part of the CI. I've set up a GitHub action to run MicroPython tests two weeks ago, and hope to get some pico-sdk examples to run as part of the CI too.

src/rp2040.ts
```diff
@@ -350,31 +372,52 @@ export class RP2040 {
...
+        this.isCore0Running = false;
+        this.core1.executeInstruction();
+      }
     }
+    this.executeTimer = setTimeout(() => this.executeInternal(), 0);
   }
 
   stop() {
-    this.stopped = true;
+    this.core0.stopped = true;
```
c1570, line 406: As far as I understand, `stopped` is a bit of a kludge for controlling the emulation machinery (that's running via `setTimeout()`) from outside code; it's not a feature of the actual RP2040 hardware. Since the cores are not running as callbacks individually, there's probably no need to give each core a `stopped` attribute.

That said, I'd just remove the `setTimeout()` stuff from rp2040js altogether; it seems to be some leftover from Wokwi. It'd probably be better to just have a `rp2040.step()` function and leave everything else to the calling code. (but this might be outside the scope of this PR)

src/rp2040.ts
```diff
@@ -350,31 +372,52 @@ export class RP2040 {
...
-    }
-    if (!this.stopped) {
-      this.executeTimer = setTimeout(() => this.execute(), 0);
+    let idle = false;
+    for (let i = 0; i < 1000 && !idle; i++) {
+      idle = true;
+      if (!this.core0.stopped && !this.core0.waiting) {
+        idle = false;
+        this.isCore0Running = true;
+        this.core0.executeInstruction();
```
c1570, line 394: This eventually gets the cores out of sync. Imagine core0 executing a number of 5 cycle instructions and core 1 executing a number of 1 cycle instructions, let things run a while, then look at the `cycles` of each core (which will have diverged a lot).

You'll need something like https://github.com/c1570/rp2040js/blob/3a754941c169d7215d8d97112c01a9d1f9d4e2c9/demo/dual-mcu.ts#L102 here.

src/rp2040.ts
```diff
@@ -48,7 +49,8 @@ export class RP2040 {
   readonly usbDPRAM = new Uint8Array(4 * KB);
   readonly usbDPRAMView = new DataView(this.usbDPRAM.buffer);
 
-  readonly core = new CortexM0Core(this);
+  readonly core0 = new CortexM0Core(this);
```
c1570, line 52: Why not using an array/list for the cores? Seems that would make quite a few places of the code simpler, and `RP4040` might be just around the corner anyways. ;)

c1570:
> Test ok with [pico-example code/multicore](https://github.com/raspberrypi/pico-examples/tree/master/multicore)

I tried with hello_multicore which gives...
```
(...the usual...)
19:16:47.164 [UART0] Unimplemented peripheral write to 24: 67
19:16:47.164 [UART0] Unimplemented peripheral write to 28: 52
19:16:47.165 [UART0] Unimplemented peripheral write to 48: 3
Hello, multicore!
19:16:47.168 [CortexM0Core] SEV
19:16:47.168 [CortexM0Core] SEV
19:16:47.168 [CortexM0Core] SEV
19:16:47.312 [RP2040] Read from invalid memory address: 20042320
19:16:47.312 [RP2040] Read from invalid memory address: 20042000
19:16:47.312 [RP2040] Read from invalid memory address: 20042000
19:16:47.312 [CortexM0Core] Warning: Instruction at 20042000 is not implemented yet!
19:16:47.312 [CortexM0Core] Opcode: 0xffff (0xffff)
```

No _It's all gone well_. Weird. Any ideas?

c1570:
If I understand correctly, the bootloader leaves core1 in waiting state, but `emulator-run.ts` jumps into crt0 immediately, bypassing the bootloader. This makes core1 reach the code at https://github.com/raspberrypi/pico-sdk/blob/f396d05f8252d4670d4ea05c8b7ac938ef0cd381/src/rp2_common/pico_standard_link/crt0.S#L315 where things go wrong (possibly because core0 is still copying things to RAM at that point).

Adding `mcu.core1.waiting = true;` to emulator-run seems to fix the issue. :)

urish:
Thanks for reviewing this @c1570! Unfortunately, I hadn't had a chance yet to review it myself. 

I got the Pico SDK CI to work: https://github.com/wokwi/rp2040js/blob/master/.github/workflows/ci-pico-sdk.yml, which can be a good starting point for adding a multicore test to the CI.

c1570:
@urish How about just adding the needed example .hex files and use those? This would make tests easier/faster to run for users and allow testing things like `-DPICO_COPY_TO_RAM` more easily.

src/rp2040.ts
mingpepe: The `stopped` flag for each core is necessary. Take [pico-debug](https://github.com/majbthrd/pico-debug) as an example, it use one core to debug another core.

src/rp2040.ts
mingpepe: Yes, the two core execution time may diverge if they run a while if `stopped` or `waiting` flag raised. Currently I do not have good solution for it. But what you did may not be helpful(if I understand correctly).

If one core is stopped or waiting, these state will not changed until the next `setTimeout` call since there is only 1 thread in our execution environment. Then your code to call `core.step` actually do nothing.

src/rp2040.ts
mingpepe: At the beginning, I use array of cores, but found so may cases need to handle for each core(like notify another core interrupt or operations no SIO) which makes the benefit of array less and make the code not easy to read(compare to use 2 variable). It's just my opionion.

mingpepe:
@1570 Where is the code to bypass bootloader in `emulator-run.ts`?

src/rp2040.ts
c1570: The pico-debug project uses the SYSCFG DBGFORCE register, basically rerouting the SWD functionality built into the (real, hardware) Cortex cores via USB. This is unrelated to `stopped`, and SWD functionality isn't implemented at all in rp2040js.

src/rp2040.ts
c1570: Something like this is needed in `cortex-m0-core.ts`:
```
  executeInstruction() {
    if (this.waiting) {
      this.cycles++;
      return;
    }
    ...
```
...and something like this in `rp2040.ts` (this is my local version with #117 already merged):
```
  step() {
    let core0StartCycles = this.core0.cycles;
    this.isCore0Running = true;
    this.core0.executeInstruction();
    this.isCore0Running = false;
    while(this.core1.cycles < this.core0.cycles) {
      this.core1.executeInstruction();
    }
    for(let cycle = core0StartCycles; cycle < this.core0.cycles; cycle++) {
      this.pio[0].step();
      this.pio[1].step();
    }
  }
```

src/rp2040.ts
mingpepe: What I want to say is the 2 cores may be 1 running and 1 stopped. And you are right, pico-debug use SWD which is not related to our `stopped` flag. But you can check gdb server in this project which only stop core0 while debugging.

c1570:
> @1570 Where is the code to bypass bootloader in `emulator-run.ts`? I did not ever use `-DPICO_COPY_TO_RAM` before.

https://github.com/wokwi/rp2040js/blob/6344182c491904aabe1051c87ec89615491e94b1/demo/emulator-run.ts#L21 jumps to 0x10000000 (the hex file contents) but the bootrom is at 0x0 basically (if I understand correctly).

mingpepe:
> > @1570 Where is the code to bypass bootloader in `emulator-run.ts`? I did not ever use `-DPICO_COPY_TO_RAM` before.
> 
> https://github.com/wokwi/rp2040js/blob/6344182c491904aabe1051c87ec89615491e94b1/demo/emulator-run.ts#L21
> 
> jumps to 0x10000000 (the hex file contents) but the bootrom is at 0x0 basically (if I understand correctly).

Thanks.

urish (changes requested):
Thanks again for working on this!

Did a few tests, ran all the multicore examples, and they seem to work well. Also did a quick review of the code, left one minor comment.

Have you looked at adding support for debugging the 2nd core using GDB?

I believe the way to go is to expose each code as a different execution thread to GDB. This is essential so in the future we'll be able to debug issues involving both cores.

mingpepe:
@urish. To support for 2nd core with GDB, I just slightly traced code. Here are some issues to deal with.

1. Which gdb server configuration do we need? 
    a) The [official config](https://github.com/raspberrypi/openocd/blob/rp2040/tcl/target/rp2040.cfg) use `smp` and `hwthread` which shows only 1 TCP port to gdb client
    b) Expose 2 TCP ports to gdb clients => Two instance of `GDBTCPServer` with different port in our case.
2. Currently timers in `RealTimeClock` do not distinguish for which core owned. When one core is stopped, I'm not sure if it's timers should keep going or should be paused (like DMA).
    
For gdb server config a), when 1 break point hit, two cores are stopped (I do not know if it can config to let another core free run). I think in this way some gdb commands currently not implemented (but I do not know the details about gdb command yet). This may be complicated compare to config b) but let me debug my code with VSCode extension which is good for me. For config b), I'm not sure if VSCode extension can achieve connecting 2 gdb server.

urish:
Thanks for looking into this! Here are my thoughts:

1. Which gdb server configuration do we need? single GDB port, multiple hw threads. 
2. From my experience, pausing everything (both cores & all timers) makes life much easier when debugging. You want to have everything as predictable as possible, so if you set a breakpoint and step over, you want to get consistent results regardless of how fast you're stepping over.

So config a), and stop everything when we hit a breakpoint (or debugger sends a break command).

mingpepe:
If 2 cores always stopped simultaneously, do we need multi-thread to implement this feature? As I understanding, one thread is enough.

urish:
You want some way to show GDB the state of each core - so it would show you the call stack of each core, and let you step the core that you are interested it.

karaketir16:
Any progress on this?

c1570:
https://github.com/c1570/rp2040js has a dual core enabled GDB interface now (along with an MCP server).

PR #113, opened 2023-01-21
fix(adc): add error flag just if an error appeared and 16 bit resolution is configured (therefore byte_shift flag for 8 bit DMA transfers is disabled)

sreimchen:
At the moment the error flag is always set (15th bit) if full resolution (12 bit) reads are configured which are stored as 2 byte samples in the ADC's FIFO. That is wrong in my opinion because the error flag should just be set if an error appeared. An error appears if the ADC concludes that there was an error after sampling.

### DIFF
--- a/src/peripherals/adc.ts
+++ b/src/peripherals/adc.ts
@@ -186,8 +186,7 @@ export class RPADC extends BasePeripheral implements Peripheral {
         value &= 0xfff; // 12 bits
         if (this.fcs & FCS_SHIFT) {
           value >>= 4;
-        }
-        if (this.fcs & FCS_ERR) {
+        } else if (error && this.fcs & FCS_ERR) {
           value |= FIFO_ERR;
         }
         this.fifo.push(value);
--- a/src/peripherals/adc.ts
+++ b/src/peripherals/adc.ts
@@ -186,7 +186,8 @@ export class RPADC extends BasePeripheral implements Peripheral {
         value &= 0xfff; // 12 bits
         if (this.fcs & FCS_SHIFT) {
           value >>= 4;
-        } else if (error && this.fcs & FCS_ERR) {
+        } 
+        if (error && this.fcs & FCS_ERR) {
           value |= FIFO_ERR;
         }
         this.fifo.push(value);


sreimchen:
Documentation says:
err_in_fifo – If enabled, bit 15 of the FIFO contains error flag for each sample
byte_shift – Shift FIFO contents to be one byte in size (for byte DMA) - enables DMA to byte buffers.

I didn't find much information about the error flag but I think it should just be set if an error appeared while reading the sample by the ADC. In the simulation that never happens at the moment which is fine. I don't know why it is called err_in_fifo in the SDK because that doesn't make sense to me: If the FIFO is full then the sample is dropped anyway... I think the naming chosen by the RP team is confusing... in my opinion it should be named like "error_adc_reading" or am I missing something important?

src/peripherals/adc.ts
```diff
@@ -186,8 +186,7 @@ export class RPADC extends BasePeripheral implements Peripheral {
         value &= 0xfff; // 12 bits
         if (this.fcs & FCS_SHIFT) {
           value >>= 4;
-        }
-        if (this.fcs & FCS_ERR) {
+        } else if (error && this.fcs & FCS_ERR) {
```
urish, line 189: Thanks! Adding `error &&` to the if condition makes sense to me, but what the motivation behind the `else`?

sreimchen: The shift flag makes sure to put just 1 byte samples into the queue. Without the else the samples were basically 2 bytes again after the shift (because of the added 15th bit sometimes). We could move the error block above the shift block and could mask (by 0xff) instead of shifting ... both makes sense to me but I didn't want to change more than necessary :)

urish: Looking at the datasheet, it doesn't see like the error bit is affected by shiting:

image

(see the 2nd line of the `ERR` bit description)

sreimchen: I didn't see that. Nice to see some more information about the error flag! Therefore you're right: Lets get rid of the `else` and merge it?

urish: Yes!

urish: Can you please amend the PR to remove the else?

sreimchen: Done, thank you :)

Issue #114, opened 2023-01-24
Allow specific image file.


BrianPugh:
Would you be interested in having the micropython image be specifiable via CLI args for the `start:micropython` script? I'm not a javascript/typescript dev, but if you're open to the idea, I think I could try and make a PR. Seems like something like `minimist` is the way to go for argument parsing?

urish:
Sounds like a useful option. What is your specific use case for this? What are you working on?

BrianPugh:
I'm using rp2040js as an integration tester for [Belay](https://github.com/BrianPugh/belay). I would like to swap between 2 images (micropython and circuitpython), so having the image be a CLI option seems like a natural choice.

urish:
Cool use case, thanks for explaining!

mingpepe:
This feature also useful for me.

BrianPugh:
resolved in #116

PR #116, opened 2023-01-29
Cli options "--image" and "--gdb"

BrianPugh:
PR for issue #114 

Adds CLI options to allows specification of the UF2 image being loaded as well as whether or not to start the GDB server. All default functionality is the same as before. Example usage:

```
npm run start:micropython -- --gdb --image=circuitpython.uf2
```

This is my first time using javascript/typescript, so I would love feedback on if anything in this PR isn't idiomatic.

EDIT: updated title/example command to match changing `--no_gdb` -> `--gdb`.

### DIFF
--- a/demo/micropython-run.ts
+++ b/demo/micropython-run.ts
@@ -5,11 +5,25 @@ import { ConsoleLogger, LogLevel } from '../src/utils/logging';
 import { bootromB1 } from './bootrom';
 import { loadUF2, loadMicropythonFlashImage } from './load-flash';
 import fs from 'fs';
+const minimist = require('minimist');
+
+const args = minimist(process.argv.slice(2), {
+  string: 'image',  // UF2 image to load; defaults to "rp2-pico-20210902-v1.17.uf2"
+  boolean: 'no_gdb',  // Do NOT start GDB server
+})
 
 const mcu = new RP2040();
 mcu.loadBootrom(bootromB1);
 mcu.logger = new ConsoleLogger(LogLevel.Error);
-loadUF2('rp2-pico-20210902-v1.17.uf2', mcu);
+
+var image_name: string;
+if(args.image){
+  image_name = args.image;
+} else {
+  image_name = 'rp2-pico-20210902-v1.17.uf2';
+}
+console.log(`Loading uf2 image ${image_name}`);
+loadUF2(image_name , mcu);
 
 if (fs.existsSync('littlefs.img')) {
   loadMicropythonFlashImage('littlefs.img', mcu);
@@ -17,8 +31,11 @@ if (fs.existsSync('littlefs.img')) {
   // https://github.com/wokwi/littlefs-wasm or https://github.com/littlefs-project/littlefs-js
 }
 
-const gdbServer = new GDBTCPServer(mcu, 3333);
-console.log(`RP2040 GDB Server ready! Listening on port ${gdbServer.port}`);
+var gdbServer: GDBTCPServer;
+if(!args.no_gdb){
+  gdbServer = new GDBTCPServer(mcu, 3333);
+  console.log(`RP2040 GDB Server ready! Listening on port ${gdbServer.port}`);
+}
 
 const cdc = new USBCDC(mcu.usbCtrl);
 cdc.onDeviceConnected = () => {
--- a/demo/micropython-run.ts
+++ b/demo/micropython-run.ts
@@ -8,22 +8,22 @@ import fs from 'fs';
 import minimist from 'minimist';
 
 const args = minimist(process.argv.slice(2), {
-  string: 'image',  // UF2 image to load; defaults to "rp2-pico-20210902-v1.17.uf2"
-  boolean: 'gdb',  // start GDB server on 3333
-})
+  string: 'image', // UF2 image to load; defaults to "rp2-pico-20210902-v1.17.uf2"
+  boolean: 'gdb', // start GDB server on 3333
+});
 
 const mcu = new RP2040();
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -61,5 +61,9 @@
     "**/*.js": [
       "prettier --write"
     ]
+  },
+  "dependencies": {
+    "@types/minimist": "^1.2.2",
+    "minimist": "^1.2.7"
```
urish, line 67: Please move these to `devDependencies`, otherwise they get installed to all users of the rp2040js library

demo/micropython-run.ts
```diff
@@ -5,20 +5,37 @@ import { ConsoleLogger, LogLevel } from '../src/utils/logging';
 import { bootromB1 } from './bootrom';
 import { loadUF2, loadMicropythonFlashImage } from './load-flash';
 import fs from 'fs';
+const minimist = require('minimist');
+
+const args = minimist(process.argv.slice(2), {
+  string: 'image',  // UF2 image to load; defaults to "rp2-pico-20210902-v1.17.uf2"
+  boolean: 'no_gdb',  // Do NOT start GDB server
```
urish, line 12: please change to `no-gdb` - it's more conventional to use dashes instead of underscores to separate words in command line arguments

demo/micropython-run.ts
```diff
@@ -5,20 +5,37 @@ import { ConsoleLogger, LogLevel } from '../src/utils/logging';
 import { bootromB1 } from './bootrom';
 import { loadUF2, loadMicropythonFlashImage } from './load-flash';
 import fs from 'fs';
+const minimist = require('minimist');
```
urish, line 8: why require? please change to use `import` as the above

demo/micropython-run.ts
```diff
@@ -5,20 +5,37 @@ import { ConsoleLogger, LogLevel } from '../src/utils/logging';
...
+  string: 'image',  // UF2 image to load; defaults to "rp2-pico-20210902-v1.17.uf2"
+  boolean: 'no_gdb',  // Do NOT start GDB server
+})
 
 const mcu = new RP2040();
 mcu.loadBootrom(bootromB1);
 mcu.logger = new ConsoleLogger(LogLevel.Error);
-loadUF2('rp2-pico-20210902-v1.17.uf2', mcu);
+
+var image_name: string;
```
urish, line 19: `var` is deprecated - please use `let` instead.

Also, JavaScript's convention is to use camel case - so please use `imageName` instead of `image_name`

demo/micropython-run.ts
```diff
@@ -5,20 +5,37 @@ import { ConsoleLogger, LogLevel } from '../src/utils/logging';
...
 
 if (fs.existsSync('littlefs.img')) {
   loadMicropythonFlashImage('littlefs.img', mcu);
   // Instead of reading from file, it would also be possible to generate the LittleFS image on-the-fly here, e.g. using
   // https://github.com/wokwi/littlefs-wasm or https://github.com/littlefs-project/littlefs-js
 }
 
-const gdbServer = new GDBTCPServer(mcu, 3333);
-console.log(`RP2040 GDB Server ready! Listening on port ${gdbServer.port}`);
+var gdbServer: GDBTCPServer;
```
urish, line 34: you can remove this line and add the word `const` before `gdbServer = ...` below

urish (changes requested):
Thanks! I left some comments.

Please also document the new options in the project's README (so other users can find out about them)

demo/micropython-run.ts
mingpepe: Argument for not starting gdb server is not intuitive for me. For my use case, I seldom use it. The reverse logic may be more intuitive.

BrianPugh: @urish which would you prefer? I like having a `--gdb` flag better, but did the opposite to maintain backwards compatibility.

urish: `--gdb`

package.json
BrianPugh: moved to `devDependencies`

demo/micropython-run.ts
BrianPugh: updated!

demo/micropython-run.ts
BrianPugh: I thought the scope of the variable might cause the server to prematurely terminate, but it seems like thats not the case 👍 .

urish:
Thanks for making the changes!

Can you please also format the code with prettier and make sure eslint is happy? 

`npx prettier --write filename` should do the trick

There's a pre-commit hook that should do this automatically, but it didn't work in your case for some reason

demo/micropython-run.ts
```diff
@@ -5,20 +5,36 @@ import { ConsoleLogger, LogLevel } from '../src/utils/logging';
...
 mcu.loadBootrom(bootromB1);
 mcu.logger = new ConsoleLogger(LogLevel.Error);
-loadUF2('rp2-pico-20210902-v1.17.uf2', mcu);
+
+let imageName: string;
+if (args.image) {
+  imageName = args.image;
+} else {
+  imageName = 'rp2-pico-20210902-v1.17.uf2';
+}
```
urish, line 24: Could be optionally shorted into:

```ts
const imageName = args.image ?? 'rp2-pico-20210902-v1.17.uf2';
```

urish:
Thanks! Then I think the only things we are missing is:

> Please also document the new options in the project's README (so other users can find out about them)

BrianPugh:
sorry I missed that! Added

urish:
Thanks!

PR #117, opened 2023-01-30
PIO timing improvements and dual MCU demo

c1570:
This makes the PIOs run in sync with the ARM cores and enables ClockDiv.
This also adds a new demo program that runs two rp2040js instances with the two RP2040s wired together using a bus.


c1570:
@urish From my side, this is ready for merging.

urish:
Update: I'm working on a new simulation clock implementation that makes it easier to keep the peripherals synchronized accurately, and will also be more GC friendly. Plan to publish it shortly under "accurate-clock" branch - stay tuned

urish:
Note: pushed the new clock implementation to the accurate clock branch - see #137

PR #119, opened 2023-02-05
fix ARM cycle counting

c1570:
add APB penalties (see datasheet "2.1.3. APB Bridge"), add missing LDR penalties

Test program: https://gist.github.com/c1570/7cd7407ea768898a7c0ada8bdbfea19e


urish:
Thanks!

PR #129, opened 2023-06-26
Report UART status as events; addl UART events

flaviut:
This is more flexible & allows for multiple listeners.

The new events allow the user to push data to the UART without a ton of trouble & setTimeouts.

This does introduce a new dependency, `eventemitter3`. This is not really needed for Node, but it is needed on the browser since the browser's events implementation is different than in Node.


flaviut:
I guess this is really two proposals in one?

- Using EventEmitter for callbacks, rather than a manually-assigned field
- Adding additional event listeners to the UART

I'd be interested to hear what you think of them, and if you think this is a good direction to go in at all.

PR #130, opened 2023-08-25
boost TX_FIFO_SIZE to 300kb

guysv:
I wondered why I can't paste into the micropython demo chunks longer than 512 bytes. then I read the code and figured out the fifo is too small, and when pasting data the mcu green thread is starving, and data starts to leak. boosting this const up allows pasting longer text easily.


urish:
Thanks! However, I think it will make more sense, in this case, to dynamically allocate/extend the FIFO based on usage. Otherwise, 300k is just as arbitrary as 512 bytes.

guysv:
yeah sure just shared it in case it bugs other people anyway.

PR #131, opened 2023-09-02
Support SPI DMA

### LINKED ISSUE #98: ILI9341 freeze with adafruit_ili9341 lib
When i'm try run this code:
```
import board
import busio
import displayio
import terminalio
import adafruit_ili9341
from adafruit_display_text import label

dc=board.GP5
rst=board.GP6
blk=board.GP7
cs=board.GP8
 
spi = busio.SPI(clock=board.GP2, MOSI=board.GP3, MISO=board.GP4)
 
displayio.release_displays()
display_bus = displayio.FourWire(spi, command = dc, chip_select=cs, reset=rst)
 
display = adafruit_ili9341.ILI9341(display_bus, width=320, height=240, rotation = 180)

splash = displayio.Group()
display.show(splash)
```
the emulator freezes and the following messages appear in the js console

image

guysv:
potentially closes #98, see comment there.

must admit this looks even slower than straight spi register to callback path... we probably want to squeeze the CPU better here.

### DIFF
--- a/.github/workflows/ci-micropython.yml
+++ b/.github/workflows/ci-micropython.yml
@@ -24,6 +24,10 @@ jobs:
         env:
           VERSION: ${{ matrix.micropython_version }}
       - name: Create filesystem
-        run: python test/mklittlefs.py
+        run: python test/mklittlefs.py littlefs.img test/micropython/main.py
       - name: Test Micropython
         run: timeout 10 npm run start:micropython -- --image micropython.uf2 --expect-text "Hello, MicroPython!"
+      - name: Create SPI test filesystem
+        run: python test/mklittlefs.py littlefs-spi.img test/micropython/main-spi.py
+      - name: Test Micropython SPI
+        run: timeout 10 npm run test:micropython-spi -- "hello world" "h" "0123456789abcdef0123456789abcdef0123456789abcdef"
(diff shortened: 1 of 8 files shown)


urish:
Thanks! Looks like there are some missing calls:

1. in doTX, after pulling an item from the txFIFO, I think we should call updateDMATx
2. in completeTransmit, after pushing an into to the rxFIFO, I think we should call updateDMARx

Also, please format the code with prettier (it should happen automatically when you commit, not sure why it was not working for you)

guysv:
i just moved the calls to updatefifos seems related to there.

guysv:
I think something broke, DMA SPI works fine but direct SPI writes x4 the data, ie every spi.write is translated into 4 register writes. very weird.

urish:
Sounds weird indeed, how did you observe that behavior?

guysv:
I added a micropython test that checks the spi call. should be ok. don't know what happened to what i found in the earlier message 🤣

urish:
Thank you!

urish:
Looks like the SPI test is failing with older releases of MicroPython (e.g. 1.19.1: https://github.com/wokwi/rp2040js/actions/runs/6247824728/job/17028285253?pr=131)

Mind looking into that?

guysv:
looks like I forgot to pass the expected SPI messages as parameters to the CI npm test invocation. please try again.

PR #145, opened 2025-04-09
Feature: additional peripheral registers

rivimey:
I tried running an instance of Apache nuttx compiled for rp2040 and got quite a number of failures, mostly due to incomplete peripheral register emulation. This PR includes some work that starts to fill in those gaps. I split the changes up by file but mostly also by function:

- Add register addrEndp with masking value; not yet used.
        AddrEndP is used to set the USB endpoint addresses for both host and device mode. nuttx was trying to configure it and failing the write, though I'm not sure it actually matters yet (other than a message) because so much usb functionality is missing anyway.

- Include register info for most registers, no additional functionality
        Adds the register names/address offsets for many of the SSI registers and for about half (those that nuttx references) adds a variable that can be read/written. I don't see any actual functionality for SSI so there doesn't seem to be anything else to do.

- Report unimplemented registers using hex values; NB sometimes see "0x-1" -- signedness issue
        This is the generic "unimplemented" code, and the change causes register offset number and register value to be written in hex - more useful IMO than decimal, and also clearer when the 0x prefix indicates it.

- Logic for spikelen register, enable_status register and sda_hold
        The I2C peripheral has a few registers that caused problems because nuttx code was busy-waiting for non-zero read results. The change makes the register model a lot more complete and can now boot nuttx[i2c].

- Add registers for all clocks in clock block with reset default values, selects, reserved bit masking
        Nuttx tries to configure the clock unit and needs to see things selected properly to boot. The change adds almost all (not the freq counter) of the registers with appropriate links, though it doesn't as yet attempt to use the clock configuration in any way. It would be nice to link the 125MHz clock constant seen elsewhere into this but I'm not sure what that would entail. Reads of control register values are anded with a bitmask corresponding to the reserved bits in the registers. Not sure if it's needed but decided it wouldn't hurt.

- Report correct state when GPIO has both pullup & pulldown
        This is a little random: saw a comment in the gpio code about both pullup&pulldown at once and decided to include a fix, based on the datasheet. I looked for possible places the new state might be used but didn't see any...

Not changed:
    The nuttx boot also tries to change PLL and XOSC unit registers but these blocks are not implemented at all and I chose not to start doing so. Perhaps in the longer term this might be done.

I do get an apparently valid boot with these changes included (using rp2040js via npm on Mac) though it is hard to know how correct the result is, given how much of the hardware is not emulated at all.

I had hoped to be able to use nuttx driving the USB unit in USB CDC-EN (Ethernet) mode, but I guess this would be a lot of work from here.

Hope this is useful. I'm up for some refinement if you need it.

### DIFF
--- a/src/gpio-pin.ts
+++ b/src/gpio-pin.ts
@@ -191,11 +191,9 @@ export class GPIOPin {
       if (this.pulldownEnabled && this.pullupEnabled) {
         // Pull high when high, pull low when low:
         return GPIOPinState.InputBusKeeper;
-      }
-      else if (this.pulldownEnabled) {
+      } else if (this.pulldownEnabled) {
         return GPIOPinState.InputPullDown;
-      }
-      else if (this.pullupEnabled) {
+      } else if (this.pullupEnabled) {
         return GPIOPinState.InputPullUp;
       }
       return GPIOPinState.Input;
--- a/src/gpio-pin.ts
+++ b/src/gpio-pin.ts
@@ -7,6 +7,7 @@ export enum GPIOPinState {
   Input,
   InputPullUp,
   InputPullDown,
+  InputBusKeeper,
 }
 
 export const FUNCTION_PWM = 4;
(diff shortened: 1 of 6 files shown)


urish:
Thanks for your contribution and for the detailed explanation! 

Can you please fix the few code styles issues? https://github.com/wokwi/rp2040js/actions/runs/14362416160/job/40271713997?pr=145

Also, would love if we had a github workflow to test with Nuttx, similar to what we currently do with the Pico SDK and MicroPython: https://github.com/wokwi/rp2040js/blob/main/.github/workflows/ci-micropython.yml

But this could be as a future PR

PR #146, opened 2025-06-04
Improve standalone emulator run cli

baileyluTCD:
Recently (in #143), the `npm start` script gained the ability to have a `--image` flag passed. I believe this functionality would be quite useful as a standalone packaged program, and hence am in the [process of packaging for nix](https://github.com/NixOS/nixpkgs/pull/413734).

To that end, this pull request implements:
- `--help` and `--version` flags for `demo/emulator-run.ts` to improve the general user experience from the command line.
- A `npm run build:demo` script with it's own `tsconfig.demo.json` for building/running demo scripts as their own programs.

### DIFF
--- a/demo/emulator-run.ts
+++ b/demo/emulator-run.ts
@@ -1,37 +1,71 @@
-import * as fs from 'fs';
-import { GDBTCPServer } from '../src/gdb/gdb-tcp-server.js';
-import { Simulator } from '../src/simulator.js';
-import { bootromB1 } from './bootrom.js';
-import { loadHex } from './intelhex.js';
-import { loadUF2 } from './load-flash.js';
-import minimist from 'minimist';
+import * as fs from "fs";
+import minimist from "minimist";
+import packageJson from "../package.json" with { type: "json" };
+import { GDBTCPServer } from "../src/gdb/gdb-tcp-server.js";
+import { RP2040 } from "../src/index.js";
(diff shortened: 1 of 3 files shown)


baileyluTCD:
> Note that there are some formatting issues with the modified code, please run prettier to fix them.

Should be fixed now, prettier did not support the `with { type: "json" }` addendum for imports hence I removed it.

> Also, note that we are not married to minimist - alternatives such as [Commander.js](https://www.npmjs.com/package/commander) or [meow](https://www.npmjs.com/package/meow) do stuff like handling --help automatically.

In light of this I switched over to [sade](https://www.npmjs.com/package/sade) which does have automatic handling of `--help` and `--version` and as a bonus actually a smaller bundle size than minimist did.

Additionally you should note that if the original `tsconfig.json` and build scripts are modified a bit you could support `npx rp2040js` with the `bin` field of the `package.json` with very little effort, I'm happy to implement this here too if desired to make it easier for others to use directly from npm.

baileyluTCD:
> Thanks for switching, and sorry about the long response times.

No worries, these things happen.

I have corrected the issue with `expect-text` by using the javascript built in `TextDecoder` object instead of parsing it manually and it seems to work in CI for every micropython version tested bar the `20210618-v1.16` version, do you have any insight as to why this may happen?

PR #150, opened 2026-03-15
Fix a bug with add rn, pc instructions

pguyot:
Fix a bug in the emulation of add rn, pc instructions. Value should be incremented by pc + 4, not pc + 2.
Confirmed against qemu.


urish:
Thanks!

urish:
Fix released as v1.3.1

PR #152, opened 2026-06-25
feat(pio): implement TX/RX FIFO join (FJOIN_TX / FJOIN_RX)

begeistert:
## Problem

The PIO state machines always expose two fixed 4-deep FIFOs and ignore the `FJOIN_TX` / `FJOIN_RX` bits in `SHIFTCTRL`. Programs that merge the two halves into a single 8-deep FIFO — a common idiom for deeper TX buffering or RX capture — observe the wrong depth, `FLEVEL`, and `FSTAT` full/empty flags.

## Change

- `FIFO` gains an adjustable usable depth (`setCapacity`) on top of its allocated buffer; `full` / `size` now honour it.
- `StateMachine` allocates 8-deep FIFOs and applies the join configuration via `updateFifoJoin()` on construction and on every `SHIFTCTRL` write. The joined direction becomes 8 deep, the donated direction is disabled (depth 0, reads back as both empty and full), and reconfiguring the join empties both FIFOs — matching the RP2040 datasheet (§3.5.4).

## Tests

Adds tests covering the 8-deep TX and RX joins, overflow / `TXOVER` behaviour, the disabled opposite direction, and FIFO clearing on reconfiguration.

---

For context: this came out of RP2040Sharp, a C# port of rp2040js, while differential-testing the PIO against real RP2040 silicon — the FIFO-join semantics here were validated against hardware. Contributing it back upstream.

### DIFF
--- a/src/peripherals/pio.spec.ts
+++ b/src/peripherals/pio.spec.ts
@@ -40,6 +40,7 @@ import {
 } from '../utils/pio-assembler.js';
 
 const CTRL = 0x50200000;
+const FSTAT = 0x50200004;
 const FLEVEL = 0x5020000c;
 const TXF0 = 0x50200010;
 const RXF0 = 0x50200020;
@@ -72,6 +73,7 @@ const RX0_SHIFT = 4;
 
 // SM0_SHIFTCTRL bits:
 const FJOIN_RX = 1 << 30;
+const FJOIN_TX = 1 << 31;
 const IN_SHIFTDIR = 1 << 18;
 const OUT_SHIFTDIR = 1 << 19;
 const SHIFTCTRL_AUTOPULL = 1 << 17;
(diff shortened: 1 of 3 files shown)


urish:
Thank you for the pull request!

I see there's one major issue that needs fixing before can we merge it:

Per pico-sdk/src/rp2040/hardware_regs/include/hardware/regs/pio.h:

```cpp
#define PIO_SM0_SHIFTCTRL_FJOIN_TX_BITS  _u(0x40000000)   // bit 30
#define PIO_SM0_SHIFTCTRL_FJOIN_RX_BITS  _u(0x80000000)   // bit 31
```

The PR has it the other way round:

```js
const SHIFTCTRL_FJOIN_RX = 1 << 30;  
const SHIFTCTRL_FJOIN_TX = 1 << 31; 
```

And one smaller thing - `updateFifoJoin()` updates the DREQs but not `checkInterrupts()`.

